PR #30, opened 2021-02-20
Add gh-action for both PR as well as for merge into master

derek-keeler:


derek-keeler:
The way I had it before was the action runs only for PRs and _not_ for merges. That's not very helpful as you kind of want both.

PR #31, opened 2021-02-20
Multiplayer status

eniklas:
Previously, we were using `max_players` to decide whether a game is multiplayer or not, which is problematic when we don't know that value. With this change, we now set a boolean `multiplayer` for each title, which is `True` if and only if:
- max players > 1 (no matter how that number was reached)
- the game modes in IGDB include a multiplayer mode

We won't filter games that we have no multiplayer info on, which I think is preferable because we're better off assuming a game is multiplayer if we know nothing about it; it's easier to catch them as the multiplayer list is much shorter, and less impactful to see a few extra games we don't care about rather than omitting games we do care about.

I also got rid of the `include_zero_players` option, as it was a complication only needed because of the previous method.


gamatrix-gog.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# TODO: copy max players to other games with the same sanitized title
```
derek-keeler, line 2: Don't bother leaving this, just make an issue!

gamatrix-gog.py
```diff
@@ -194,21 +198,27 @@ def build_config(args):
...
 
         if "max_players" in game_list[k]:
-            log.debug(
-                f'{k}: max players {game_list[k]["max_players"]} from config file'
-            )
-            continue
+            max_players = game_list[k]["max_players"]
+            reason = "from config file"
+            if max_players > 1:
+                multiplayer = True
```
derek-keeler, line 221:
```suggestion
            multiplayer = max_players > 1
```

derek-keeler (approved):
LGTM

gamatrix-gog.py
eniklas: Yeah, was being lazy 😞 About to do that next, will remove on the next PR.

gamatrix-gog.py
eniklas: Oh yeah, nice one

PR #40, opened 2021-02-22
Interim change to switch argparse out for docopt

derek-keeler:
**NOTE**: Interim PR - not intended to merge.

This is to prove to myself and to others that it will work without breaking everything.

Do have a look and point out where I might be missing important details!

Left all the `argparse` stuff in and added the `docopt` stuff alongside it. Then tested the outcome of `build_config` for each path to compare.

You can see the doc in the header of `gamatrix-gog.py` which essentially is exactly the output you want from `--help`.

The tests essentially pass in simulated command lines (as `sys.argv` lists) and then build a config from the options recieved. Then the fields within the config are compared between old and new.

### DIFF
--- a/gamatrix-gog.py
+++ b/gamatrix-gog.py
@@ -3,7 +3,7 @@ import argparse
 import logging
 import os
 import sys
-from typing import Any, List
+from typing import Any, Dict, List
 
 from flask import Flask, render_template, request
 from ipaddress import IPv4Network
--- a/gamatrix-gog.py
+++ b/gamatrix-gog.py
@@ -3,7 +3,7 @@ import argparse
 import logging
 import os
 import sys
-from typing import Any, Dict, List
+from typing import Any, List
 
 from flask import Flask, render_template, request
 from ipaddress import IPv4Network
--- a/gamatrix-gog.py
+++ b/gamatrix-gog.py
@@ -6,12 +6,12 @@ Show and compare between games owned by multiple users.
 Usage:
     gamatrix-gog.py --help
     gamatrix-gog.py --version
-    gamatrix-gog.py [--config-file=CFG] [--debug] [--all-games] [--interface=IFC] [--include-single-player] [--port=PORT] [--server] [--userid=UID ...] [--include-zero-players] [<db> ... ]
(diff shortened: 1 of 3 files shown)


gamatrix-gog.py
```diff
@@ -103,7 +130,102 @@ def init_opts():
     }
 
 
-def build_config(args):
+def build_config(args: Dict[str, Any]) -> Dict[str, Any]:
```
eniklas, line 133: I understand generally how the `typing` stuff works, but what does `Any` mean here?

gamatrix-gog.py
```diff
@@ -103,7 +130,102 @@ def init_opts():
...
+
+    # Command-line args override values in the config file
+    # TODO: maybe we can do this directly in argparse, or otherwise better
+
+    # This can't be given as an argument as it wouldn't make much sense;
+    #  provide a sane default if it's missing from the config file
+    if "db_path" not in config:
+        config["db_path"] = "."
+
+    config["all_games"] = args.get("--all-games", False)
```
eniklas, line 158: So `docopt` names the args with the long name? Will have to get used to that, argparse and click would call that `args.all_games`.

gamatrix-gog.py
```diff
@@ -103,7 +130,102 @@ def init_opts():
...
+
+    config["include_single_player"] = args.get("--include-single-player", False)
+
+    if args.get("--server", True):  # Note that the --server opt is False unless present
```
eniklas, line 162: So you're defaulting to server mode? Not sure I quite understand how this works. If `--server` isn't used, then it defaults to `True`, so mode is set to `"server"`?

requirements.txt
```diff
@@ -1,10 +1,8 @@
...
-itsdangerous==1.1.0
 Jinja2==2.11.2
 MarkupSafe==1.1.1
 pyyaml==5.3.1
 requests==2.25.1
 ruamel.yaml==0.16.12
-ruamel.yaml.clib==0.2.2
-Werkzeug==1.0.1
\ No newline at end of file
+ruamel.yaml.clib==0.2.2
```
eniklas, line 8: Curious how you updated this. I know you typically do `pip freeze > requirements.txt`, but I believe that includes everything installed in your venv, which might include stuff that's not used in the code.

eniklas (approved):
Left a few questions for my own edumacation, but LGTM!

PR #43, opened 2021-02-24
first stab at marking cache dirty

eniklas:


helpers/cache_helper.py
```diff
@@ -14,3 +14,14 @@ def __init__(self, cache_file):
...
             json.dump(self.data, f)
+        self.data.dirty = False
+
+
+# Use this for the cache so we can mark it dirty on any update
+class CacheDict(dict):
+    def __init__(self):
+        self.dirty = False
+
+    def __setitem__(self, k, v):
```
derek-keeler: This the only place a `dict` can be updated? What about calls to `update()` or such things, does that all end up going through `__setitem__`?

gamatrix-gog.py
```diff
@@ -331,10 +335,12 @@ def parse_cmdline(argv: List[str]) -> Any:
...
 
     cache = Cache(config["cache"])
+    log.info(f"DEBUG: cache state after init: {cache.data.dirty}")
     # Get multiplayer info from IGDB and save it to the cache
     igdb = IGDBHelper(
         config["igdb_client_id"], config["igdb_client_secret"], cache.data
     )
+    log.info(f"DEBUG: cache state after igdb init: {cache.data.dirty}")
 
     if "mode" in config and config["mode"] == "server":
```
derek-keeler: Would this work? (I know this isn't what this PR is about... sorry 😙)

```suggestion
    if config.get("mode", "no") == "server":
```

gamatrix-gog.py
```diff
@@ -361,7 +367,11 @@ def parse_cmdline(argv: List[str]) -> Any:
         igdb.get_game_info(release_key)
         igdb.get_multiplayer_info(release_key)
 
-    cache.save()
+    log.info(f"DEBUG: cache state before save: {cache.data.dirty}")
+    if cache.data.dirty:
```
derek-keeler: If you make `Cache.save` check for `data.dirty` internally, you won't have to make this check anywhere outside the class, and if  you call it too often it won't matter.

gamatrix-gog.py
```diff
@@ -71,6 +71,10 @@ def compare_libraries():
         igdb.get_game_info(release_key)
         igdb.get_multiplayer_info(release_key)
 
+    # Save the cache to disk if we added anything
+    if cache.data.dirty:
```
derek-keeler: Consider adding function/property `Cache.is_dirty` and checking the `CacheDict.dirty` from within that module - then no-one else needs to know about your internal implementation of the cache.

derek-keeler (commented):
Nice work! Looking forward to seeing it in action.

helpers/cache_helper.py
```diff
@@ -1,16 +1,34 @@
...
+        self.dirty = True
+        super(CacheDict, self).__setitem__(k, v)
```
derek-keeler, line 51: Wouldn't this be `super(dict, self).__setitem__(k,v)` ?

derek-keeler: Ah, you can just write this as `super().__setitem__(k,v)` but it seems it is the same as you have it here.

eniklas:
I gave up on this for now, opened https://github.com/eniklas/gamatrix-gog/issues/45 to take care of it down the road.

PR #71, opened 2021-03-07
Switch commandline handling over to docopt

derek-keeler:
- Add tests for command line -> config
- Switch from argparse to docopt
- Small tweaks to the code

### DIFF
--- a/gamatrix-gog.py
+++ b/gamatrix-gog.py
@@ -6,7 +6,7 @@ Show and compare between games owned by multiple users.
 Usage:
     gamatrix-gog.py --help
     gamatrix-gog.py --version
-    gamatrix-gog.py [--config-file=CFG] [--debug] [--all-games] [--interface=IFC] [--include-single-player] [--port=PORT] [--server] [--userid=UID ...] [<db> ... ]
+    gamatrix-gog.py [--config-file=CFG] [--debug] [--all-games] [--interface=IFC] [--include-single-player] [--port=PORT] [--server] [--update-cache] [--userid=UID ...] [<db> ... ]
 
 Options:
   -h, --help                   show this help message and exit
@@ -18,6 +18,7 @@ Options:
   -I, --include-single-player  include single player games
   -p PORT, --port=PORT         the network port to use if running in server mode; default is 8080.
   -s, --server                 run in server mode
+  -U, --update-cache           Force a cache-update even when the cache is not dirty.
   -u USERID, --userid=USERID   the GOG user IDs to compare, there can be multiples of this switch
 
 Positional Arguments:
@@ -27,7 +28,7 @@ Positional Arguments:
 import logging
 import os
 import sys
-from ipaddress import IPv4Network
+from ipaddress import IPv4Address, IPv4Network
 from typing import Any, Dict, List
 
 import docopt
@@ -35,8 +36,8 @@ from flask import Flask, render_template, request
 from ruamel.yaml import YAML
 from werkzeug.utils import secure_filename
 
-from helpers.cache_helper import Cache
 from helpers import constants
+from helpers.cache_helper import Cache
 from helpers.gogdb_helper import gogDB
 from helpers.igdb_helper import IGDBHelper
 from helpers.misc_helper import sanitize_title
@@ -54,7 +55,7 @@ def root():
         "index.html",
         users=config["users"],
         uploads_enabled=config["uploads_enabled"],
-        platforms=PLATFORMS,
+        platforms=constants.PLATFORMS,
         version=VERSION,
     )
 
@@ -178,7 +179,7 @@ def compare_libraries():
         users=opts["user_ids_to_compare"],
         caption=gog.get_caption(len(common_games)),
         show_keys=opts["show_keys"],
-        platforms=PLATFORMS,
+        platforms=constants.PLATFORMS,
     )
 
 
@@ -360,15 +361,17 @@ def build_config(args: Dict[str, Any]) -> Dict[str, Any]:
     if "hidden" not in config:
         config["hidden"] = []
 
+    config["update_cache"] = False
+    if args.get("--update-cache", False):
+        config["update_cache"] = True
+
     # Lowercase and remove non-alphanumeric characters for better matching
     for i in range(len(config["hidden"])):
-        config["hidden"][i] = constants.ALPHANUM_PATTERN.sub(
-            "", config["hidden"][i]
-        ).lower()
+        config["hidden"][i] = sanitize_title(config["hidden"][i])
 
     sanitized_metadata = {}
     for title in config["metadata"]:
-        sanitized_title = constants.ALPHANUM_PATTERN.sub("", title).lower()
+        sanitized_title = sanitize_title(title)
         sanitized_metadata[sanitized_title] = config["metadata"][title]
 
     config["metadata"] = sanitized_metadata
--- a/gamatrix-gog.py
+++ b/gamatrix-gog.py
@@ -281,14 +281,14 @@ def build_config(args: Dict[str, Any]) -> Dict[str, Any]:
             "{}/{}".format(config["db_path"], config["users"][userid]["db"])
         )
 
-    # Convert CIDRs into IPv4Network objects; if there are none, disable uploads
-    config["uploads_enabled"] = False
-    if "cidrs" in config["users"][userid]:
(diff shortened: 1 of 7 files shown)


derek-keeler:
Apologies for such a massive PR. I tried to keep it simple, most of the change is switching from the `argparse` over to `docopt`. One bugfix for the `cli` scenario in the helpers where we are using `.keys()` on a `List` (it may have previously been a `Dict`?)

gamatrix-gog.py
```diff
@@ -228,24 +245,20 @@ def build_config(args):
     if "db_path" not in config:
         config["db_path"] = "."
 
-    config["all_games"] = False
-    if args.all_games:
-        config["all_games"] = True
+    config["all_games"] = args.get("--all-games", False)
```
eniklas, line 248: Nice!

gamatrix-gog.py
```diff
@@ -228,24 +245,20 @@ def build_config(args):
...
 
-    config["include_single_player"] = False
-    if args.include_single_player:
-        config["include_single_player"] = True
+    config["include_single_player"] = args.get("--include-single-player", False)
 
-    if args.server:
+    if args.get("--server", True):  # Note that the --server opt is False unless present
```
eniklas: `--server` is always set if I'm understanding the comment, so why set the default value?

gamatrix-gog.py
```diff
@@ -265,50 +278,34 @@ def build_config(args):
         config["db_list"].append(
             "{}/{}".format(config["db_path"], config["users"][userid]["db"])
         )
-        # Convert CIDRs into IPv4Network objects; if there are none, disable uploads
```
eniklas, line 268: Where did this block go?

gamatrix-gog.py
```diff
@@ -265,50 +278,34 @@ def build_config(args):
...
+        ):
+            config["db_list"].append(
                 "{}/{}".format(config["db_path"], config["users"][userid]["db"])
-                not in config["db_list"]
-            ):
-                config["db_list"].append(
-                    "{}/{}".format(config["db_path"], config["users"][userid]["db"])
-                )
-
-    # Order users by username to avoid having to do it in the templates
```
eniklas, line 299: Samesies, where did this go?

gamatrix-gog.py
```diff
@@ -337,7 +334,6 @@ def set_multiplayer_status(game_list, cache):
       - IGDB game modes includes a multiplayer mode
     """
     for k in game_list:
-        igdb_key = game_list[k]["igdb_key"]
```
eniklas: Looks like I hosed you with my merge last night. You'll need to merge `master` in.

helpers/gogdb_helper.py
```diff
@@ -385,7 +385,7 @@ def get_caption(self, num_games):
 
         if self.config["all_games"]:
             caption_middle = "total games owned by"
-        elif len(self.config["user_ids_to_compare"].keys()) == 1:
+        elif len(self.config["user_ids_to_compare"]) == 1:
```
eniklas, line 388: Nice catch, not sure why that hasn't been barfing.

requirements.txt
```diff
@@ -1,10 +1,8 @@
...
+docopt
 Flask==1.1.2
-itsdangerous==1.1.0
 Jinja2==2.11.2
 MarkupSafe==1.1.1
 pyyaml==5.3.1
 requests==2.25.1
 ruamel.yaml==0.16.12
-ruamel.yaml.clib==0.2.2
-Werkzeug==1.0.1
```
eniklas, line 10: How did you generate this? Pretty darn sure `Werkzeug` is needed.

eniklas (changes requested):
Needs `master` merged in, and I have a few questions about blocks that look to have been removed.

gamatrix-gog.py
derek-keeler: Ah crap! This should read `.get("--server", False)` based on the comment, and also based on the behaviour we want. I'll change it.

gamatrix-gog.py
derek-keeler: I missed it in my `git --rebase` flow, apologies! Yeah I really hate it when I let PRs get too big like this. Thank you for catching it.

requirements.txt
derek-keeler: Werkzeug is pulled in as part of another package. Do we need to lock it?

helpers/gogdb_helper.py
derek-keeler: It did barf for me!

derek-keeler:
Yep, had to apply some changes from last night! You must have been geeking out on this later than I!! :smiley: 

I started to go over your notes this morning, but am being interrupted by the fam. Will come back to it later after I've adequately distracted everyone.

gamatrix-gog.py
derek-keeler: Will go over my entire change more closely later today. I don't want to miss anything!

requirements.txt
eniklas: Ah ok, if flask is pulling it in, all good. Still curious how you generated the new requirements though. `pip freeze > requirements.txt` is the only way I know, but pretty sure that adds everything installed in your venv, whether it's used in the code or not.

eniklas:
Yeah man, I had no idea you were working on this stuff last night too. I merged mine and cut a new release, then just as I was logging off for bed I saw your PR come through!

requirements.txt
derek-keeler: Yep, that's exactly correct!

I was wondering why all this was in here, I thought maybe you had made use of `pip-tools` or something.

Instead of capturing the environment, I just add the libs I want/need and let the `pip` tool sort out dependencies. When I get bitten by version issues I tend to lock down versions only at that time. This might be a bit cavalier of me though.

derek-keeler: This way the `Werkzeug` lib will come in at the latest non-breaking version that is depended on by `Flask`, which I think is better, and not as flippant as I would have it?

derek-keeler (commented):
Sorry for the noise at the top of the file, below the `docopt` __doc__ string. I couldn't help myself from hitting the `Sort Imports` in VSCode... 🙄

I also added code coverage to this, setting the minimum

gamatrix-gog.py
```diff
@@ -468,24 +434,22 @@ def parse_cmdline(argv: List[str]) -> Any:
...
     # init_opts() is meant for server mode; any CLI options that are also
     # web UI options need to be overridden
-    opts = init_opts()
-    opts["include_single_player"] = args.include_single_player
-    for userid in user_ids_to_compare:
-        opts["user_ids_to_compare"][userid] = config["users"][userid]
-
+    web_opts = init_opts()
+    web_opts["include_single_player"] = opts.get("--include-single-player", False)
+    web_opts["user_ids_to_compare"] = user_ids_to_compare
```
eniklas, line 449: Are you sure this works? I think we still need the `for` loop on old lines 484-485 to add the data under each userid.

gamatrix-gog.py
```diff
@@ -369,91 +380,46 @@ def set_multiplayer_status(game_list, cache):
...
+                    if mode in constants.IGDB_MULTIPLAYER_GAME_MODES:
                         multiplayer = True
                         reason = f"as game modes includes {mode}"
                         break
 
         log.debug(
-            "{} ({}, IGDB key {}): multiplayer {}, max players {} {}".format(
-                k, game_list[k]["title"], igdb_key, multiplayer, max_players, reason
-            )
+            f"{k} ( {game_list[k]['title']}, IGDB key {igdb_key}): ",
```
eniklas: Now you're gonna see what a pedant I am: there's a space after the `(` that will make my eyelid twitch :-)

requirements.txt
eniklas: I do prefer to lock all versions, otherwise shit can break suddenly when you build, and you'll probably spend some time figuring out it wasn't your code changes that broke the shit.

eniklas (changes requested):
Only one thing I'm actually concerned about, the other two are nits.

gamatrix-gog.py
derek-keeler: I agree, you caught another one. :thinking: If only there was some way we could _test for such stuff_.

eniklas (approved):
👍 LGTM, I'll let you do the honours!

PR #84, opened 2021-04-11
Remove python-libmagic dependency

derek-keeler:


helpers/gogdb_helper.py
```diff
@@ -2,11 +2,33 @@
...
+def is_sqlite3(path: str) -> bool:
+    """Fast check to ensure a file is indeed an SQLite3 db file, based on the format."""
+    # Thanks! https://stackoverflow.com/a/15355790/895739
+
+    test_sql_file = pathlib.Path(path)
+
+    if not test_sql_file.exists():
+        return False
+    if not test_sql_file.is_file():
+        return False
```
eniklas, line 21: `is_file` will be `False` if the file doesn't exist, so no need for the `exists` test unless I'm missing something.

```
>>> os.path.isfile("poo")
False
```

eniklas (changes requested):
Just a nit about an unnecessary (I think) test, and you should remove [libmagic from the container](https://github.com/eniklas/gamatrix-gog/blob/master/Dockerfile#L5). And yay tests!

helpers/gogdb_helper.py
derek-keeler: What if it is a directory...

eniklas: Then `is_file()` will be `False` as well. I'm saying to get rid of `exists()`, not `is_file()`.

derek-keeler: Ahhh... smart.

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.9-slim
 
 WORKDIR /usr/src/app
 
-RUN apt update && apt install -y libmagic1 && rm -rf /var/lib/apt/lists/*
+RUN apt update && rm -rf /var/lib/apt/lists/*
```
eniklas: Just remove this whole line, it's not needed now.

eniklas (changes requested):
Almost there, just need to ditch that whole `apt` line.

Dockerfile
derek-keeler: What you don't want to just randomly update the database and then delete it for no reason!? Don't you understand who I work for?

eniklas (approved):
Nice one dude, merge away!

PR #87, opened 2021-06-06
fix sqlite check

eniklas:
File uploads were failing with

```
TypeError: expected str, bytes or os.PathLike object, not FileStorage
```

because `is_sqlite3()` was expecting a file to be passed in, but was being called with a [FileStorage](https://tedboy.github.io/flask/generated/generated/werkzeug.FileStorage.html#werkzeug.FileStorage) object. We don't have an on-disk file at the point we call this, just an object in memory. I updated the function to expect a byte stream.


dev-requirements.txt
```diff
@@ -1,5 +1,5 @@
-black
-flake8
-mypy
-pytest
-pytest-cov
+black==21.5b2
+flake8==3.8.4
+mypy==0.812
+pytest==6.2.3
+pytest-cov==2.12.1
```
eniklas, line 5: I was getting burned because I had a slightly older version of `black` locally, so it wasn't formatting the same way the latest version was and tests were failing. This is why I'm a big fan of pinning!

PR #98, opened 2021-11-14
Remove functionality of adding db in cmdline mode

### LINKED ISSUE #35: Adding DB files on the command line doesn't work
Adding db files at the end of the command line causes errors in the current `argparse` mechanism.

Not sure if this is moot or not at this point, should we remove it?

derek-keeler:
Fixes #35


eniklas (approved):
👍

PR #104, opened 2022-01-22
update DB query for new schema

eniklas:
This fixes the DB queries to use the new schema, fixing https://github.com/eniklas/gamatrix/issues/103.


derek-keeler (commented):
Easy peasy lemon squeezy.

Nice that you didn't have to alter too much!

PR #109, opened 2022-10-29
Update README.md

Graybriel:
added the syntax for using powershell as compared to having to download curl for windows


eniklas:
@Graybriel when I put this in a `.bat` file and run it, I get:

```
Invoke-WebRequest : The process cannot access the file 'C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db' because
it is being used by another process.
At line:1 char:1
+ Invoke-WebRequest -Method Post -Uri http://<my-url>/compa ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Invoke-WebRequest], IOException
    + FullyQualifiedErrorId : System.IO.IOException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand
```

Is there a way to force it to read an open file?

derek-keeler:
> @Graybriel when I put this in a `.bat` file and run it, I get:
> 
> ...
>
> Is there a way to force it to read an open file?

Try adding `-UseBasicParsing` as I suspect the IE rules underlying the Invoke-WebRequest is being a prude.

Or, alternatively, try using ` Invoke-RestMethod -Method Post -Uri "https://<your-gamatrix-url>/compare?option=upload" -InFile "C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db"` which may get around those restrictions for other reasons.

derek-keeler:
It hurts me deeply, in my heart, that you put this into a `.bat` file and not just running it from a `pwsh` shell directly in the new Windows Terminal program.

We make these things for you. Why won't you let us love you?

eniklas:
> > @Graybriel when I put this in a `.bat` file and run it, I get:
> > ...
> > Is there a way to force it to read an open file?
> 
> Try adding `-UseBasicParsing` as I suspect the IE rules underlying the Invoke-WebRequest is being a prude.
> 
> Or, alternatively, try using ` Invoke-RestMethod -Method Post -Uri "https://<your-gamatrix-url>/compare?option=upload" -InFile "C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db"` which may get around those restrictions for other reasons.

No luck with either of those options.

> It hurts me deeply, in my heart, that you put this into a .bat file and not just running it from a pwsh shell directly in the new Windows Terminal program.
>
> We make these things for you. Why won't you let us love you?

I guess because I don't know how to do that. Is that done by using another extension, e.g. `ps1`? And, will that incur restrictions that a `.bat` file doesn't, as happened when I originally set this up? Back then, I couldn't get a `.ps1` file to execute without a bunch of red tape the user would have to go through every time they ran it.

derek-keeler:
> > > @Graybriel when I put this in a `.bat` file and run it, I get:
> > > ...
> > > Is there a way to force it to read an open file?
> > 
> > 
> > Try adding `-UseBasicParsing` as I suspect the IE rules underlying the Invoke-WebRequest is being a prude.
> > Or, alternatively, try using ` Invoke-RestMethod -Method Post -Uri "https://<your-gamatrix-url>/compare?option=upload" -InFile "C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db"` which may get around those restrictions for other reasons.
> 
> No luck with either of those options.
> 
> > It hurts me deeply, in my heart, that you put this into a .bat file and not just running it from a pwsh shell directly in the new Windows Terminal program.
> > We make these things for you. Why won't you let us love you?
> 
> I guess because I don't know how to do that. Is that done by using another extension, e.g. `ps1`? And, will that incur restrictions that a `.bat` file doesn't, as happened when I originally set this up? Back then, I couldn't get a `.ps1` file to execute without a bunch of red tape the user would have to go through every time they ran it.

I've tried multiple attempts at reading a file when another process has "locked" it and failed (in Windows). The fact  seems to hold that if the calling process says to Windows "do not share" then [Windows listens](https://learn.microsoft.com/en-us/dotnet/api/system.io.fileshare?view=net-6.0&viewFallbackFrom=windowsdesktop-6.0). I cannot find Win32 or other docs supporting this theory though.

However:

If I do this in an Admistrator PowerShell terminal:

```pwsh
> Set-ExecutionPolicy -Scope CurrentUser Default
> Invoke-WebRequest -Method Post -Uri "http://games.erikniklas.net/compare?option=upload" -InFile "C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db"
```
It works! (When GOG is not running!)


When I run:
```pwsh
> Set-ExecutionPolicy -Scope CurrentUser Default
> Invoke-WebRequest -Method Post -Uri "http://games.erikniklas.net/compare?option=upload" -InFile "C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db"
```
It fails with the same error you have outlined above, when _GOG is running_.

Try running this yourself when GOG is not running (ensure  by checking task manager in Windows) and let me know if that is the case and I'll  suggest  an enhance set of docs.

derek-keeler:
See this discussion regarding Python in general to get an [idea how `PowerShell` might get in your way a bit](https://code.visualstudio.com/docs/python/python-tutorial#_install-and-use-packages). See the part on "1. Create and activate the virtual environment" for details...

Graybriel:
copying the file first, works while GOG is running, but it's not elegant;

if (-not (Test-Path -Path "C:\ProgramData\GOG.com\Galaxy\storage\gamatrix")) {
    mkdir -Path "C:\ProgramData\GOG.com\Galaxy\storage\gamatrix"
}

Copy-Item -Path "C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db" -Destination "C:\ProgramData\GOG.com\Galaxy\storage\gamatrix\galaxy-2.0.db" -Force
Invoke-WebRequest -Method Post -Uri "http://<<URL>>/compare?option=upload" -InFile "C:\ProgramData\GOG.com\Galaxy\storage\gamatrix\galaxy-2.0.db"

eniklas:
> copying the file first, works while GOG is running, but it's not elegant;
> 
> if (-not (Test-Path -Path "C:\ProgramData\GOG.com\Galaxy\storage\gamatrix")) { mkdir -Path "C:\ProgramData\GOG.com\Galaxy\storage\gamatrix" }
> 
> Copy-Item -Path "C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db" -Destination "C:\ProgramData\GOG.com\Galaxy\storage\gamatrix\galaxy-2.0.db" -Force
> Invoke-WebRequest -Method Post -Uri "http://<>/compare?option=upload" -InFile "C:\ProgramData\GOG.com\Galaxy\storage\gamatrix\galaxy-2.0.db"

So `Copy-Item` is fine reading the open file, more evidence that it's a limitation of `Invoke-WebRequest`.

eniklas:
Per our discussion on Discord:
- curl is available in Windows by default, no need to install it
- when using powershell, `curl` is aliased to `Invoke-WebRequest`, which won't work as it has different arguments; using `curl.exe` uses the actual curl
- we weren't able to find a way to get `Invoke-WebRequest` to work with an open file

I updated the PR to remove the powershell option, and removed the instruction to download curl.

@Graybriel I think a `.ps1` file would look exactly like the `.bat` file, except it must use `curl.exe` to work, so I just left the `.bat` file in there. If you know of an advantage to using a `.ps1` file I'm good to put it back in.

PR #111, opened 2022-11-14
Modernize Python project layout & setup

derek-keeler:
Switch to using the more modern project setup for Python.

- Use `pyproject.toml` to define build/test/project metadata and dependencies.
  - Update source layout to work with _setuptools_ `src-layout` scheme (as opposed to `flat-layout`).
- Produce packages from a 'build' that can be used to deploy.
  - Update ci to generate deployable packages.
- Alter docker build setup to use a `whl` package and to reduce image size.


derek-keeler:
Note: @eniklas you can try out my changes via the new uploaded `wheel` file [the updated PR flow will produce](https://github.com/eniklas/gamatrix/actions/runs/3457904348). I was  thinking we could update your release workflow with that, too.

derek-keeler:
Note the new Contributing section at the bottom of the README - it's got the instructions for how to install your project dependencies.

src/gamatrix/__main__.py
```diff
@@ -117,7 +117,7 @@ def upload_file():
         <!doctype html>
         <title>Upload DB</title>
         <h1>Upload DB</h1>
-        GOG DBs are usually in C:\ProgramData\GOG.com\Galaxy\storage\galaxy-2.0.db
+        GOG DBs are usually in C:\\ProgramData\\GOG.com\\Galaxy\\storage\\galaxy-2.0.db
```
eniklas, line 120: Why the double slashes?

derek-keeler: Warnings in a run of `mypy` IIRC.

derek-keeler: No I was mistaken. The new Intellisense engine for VSCode-Python was the source of the warning. `Pylance`.

image

derek-keeler: (the other two are due to use of `or` in an atomic non-evaluated line of code)

.flake8
```diff
@@ -0,0 +1,7 @@
+# Note: Flake8 does not support pyproject.toml
+
+[flake8]
+ignore = C,E,W
```
eniklas: What are we ignoring here and why? Worth a comment here.

Dockerfile
```diff
@@ -1,11 +1,19 @@
-FROM python:3.9-slim
+FROM python:3.11-slim
 
 WORKDIR /usr/src/app
 
-COPY requirements.txt ./
-RUN pip install --no-cache-dir -r requirements.txt
+# Limit what we copy to keep image size down.
+# We currently only need the src/ folder and the pyproject.toml file.
```
eniklas: [NIT] I'm against using time-based words like "currently" in comments and docs, as they can quickly become misleading or, as in this case, simply redundant.

README.md
```diff
@@ -273,4 +273,59 @@ Now you can open the port on your router. For more information on using iptables
...
+. .venv/bin/activate        # Linux/MacOS
+.venv/Scripts/Activate.ps1  # Windows
+
+python -m pip install -U pip
+python -m pip install .[dev]
+```
+
+#### Details
+
+1. Install Python 3.7 or above. Currently, the project should support 3.7 through to the latest version.
```
eniklas, line 282: [NIT] Would remove the second sentence for the same reason as before.

README.md
```diff
@@ -273,4 +273,59 @@ Now you can open the port on your router. For more information on using iptables
...
+1. Fork the gamatrix repository into your personal profile using the `Fork` button on GitHub.com.
+1. Sync to the sources. ( `git clone https://github.com/my-github-username/gamatrix` ) and `cd` into your clone ( `cd gamatrix` ).
+1. Set up a virtual environment for managing Python dependencies.
+    - Linux/MacOS ( `python -m venv .venv` ) # assumes your system Python points at 3.7+
+    - Windows ( `py -m venv .venv` )         # sets up the latest Python version on your system
```
eniklas: IMO a more readable format would be:

- Linux/MacOS: `python -m venv .venv # assumes your system Python points at 3.7+`

Note that multiple spaces in markdown are collapsed into a single space, so the comments as is won't line up in the rendered page.

README.md
```diff
@@ -273,4 +273,59 @@ Now you can open the port on your router. For more information on using iptables
...
+development environment is set up to generate the `whl` package file. Note that the CI
+system (GitHub) will also generate the `whl` file you create with your GH Actions if you
+set them up.
+
+```bash
+python -m pip install .[ci]     # install the build tools
+python -m build --wheel         # generate the dist/gamatrix-[ver]-none-any.whl file
+```
+
+For building the Docker image you will see that the [`Dockerfile`](./Dockerfile) now generates
```
eniklas: [NIT] Would remove "now".

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+    {name = "Erik Niklas", email = "github@bobanddoug.com"},
+    {name = "Derek Keeler", email = "34773432+derek-keeler@users.noreply.github.com"},
+    {name = "Klosteinmann", email = "34807323+klosteinmann@users.noreply.github.com"},
+]
+description = """A tool to compare the games owned by several users, and list
+                 all the games they have in common. It requires all users to
+                 use [GOG Galaxy](https://www.gog.com/galaxy); since GOG Galaxy
+                 supports almost all major digital distribution platforms."""
+readme = "README.md"
+requires-python = ">=3.9"
```
eniklas: This should be `3.7`. I'm running 3.8 and it works fine if I change this.

pyproject.toml
```diff
@@ -0,0 +1,66 @@
+[project]
+name = "gamatrix"
+version = "1.4.3"
```
eniklas, line 3: Can this be gotten from `version.txt`, or can we get rid of that file? It sucks to have to remember to update the version in two places.

README.md
```diff
@@ -273,4 +273,59 @@ Now you can open the port on your router. For more information on using iptables
...
+If you do wish to contribute, here's a quickstart to get your development environment up and running:
+
+#### Quick Start
+
+__Linux/MacOS__
+
+```bash
+git clone https://github.com/my-github-username/gamatrix
+cd gamatrix
+python -m venv .venv
```
eniklas: Any objection to making this `python3`? That's what it is on my machine, and I believe most distros would have that.

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "License :: OSI Approved :: MIT License",
+    "Natural Language :: English",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
eniklas, line 27: I don't really understand this, but I assume 3.7 and 3.8 should be added.

src/gamatrix/__main__.py
eniklas: Ok, we _could_ tell Pylance to ignore the line since it's not Python. But this is fine; I checked and it still renders with single backslashes so no harm.

eniklas (changes requested):
Looks great, just a few small things. I have the nickname The Nit King in some circles...

I would remove the TOC at the top of the readme. Your vscode didn't update it, but we can get rid of it since the TOC is built into GitHub now.

README.md
eniklas: And, as mentioned above, if you just make this `python3` you can dispense with the comment altogether.

src/gamatrix/__main__.py
derek-keeler: I tried to use `# type: ignore` to get around this (and tell `Pylance` to shush here) but apparently it didn't respect the multiline string layout (I have to separate out the one line with backslashes, gross). So I'm going with double-backslash. Let's all just blame Microsoft for this one all up and be done with it.

.flake8
derek-keeler: Flake8 is gone. It doesn't support `pyproject.toml`. Also, we were basically telling it to shush anyways 😑.

README.md
derek-keeler: Agreed. Updated both Windows (to `py -3...`) and Linux/Mac (to `python3 ...`).

pyproject.toml
derek-keeler: Agreed. Created #116 for me to come back and sort this out. I have a plan! (Now that the project generates `whl` packages we can automate it during release, and have a static 'dev' version that we use when we build locally).

README.md
derek-keeler: Yes, that works well. I'll change that. (I use a complicated `pyenv`  setup so I'm usually not using `python3` anymore, but that works with `pyenv` too!

pyproject.toml
derek-keeler: [Classifiers in PyPI](https://pypi.org/classifiers/) - I did steal this part from another project that does release to PyPI and even though we don't (and maybe never will), I thought I'd leave it. I'm not going to lose any sleep over it if you want it gone, total [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

For now, I've updated it to include 3.7 & 3.8.

eniklas (approved):

derek-keeler:
🎆

PR #118, opened 2023-02-11
Added dev-workflow documentation

derek-keeler:
- made note of the pip install -e switch


eniklas (approved):
Nice one, thanks for digging into that! ❤️

PR #121, opened 2023-02-12
get version from pyproject.toml

eniklas:
This removes `version.py` and gets the version from `pyproject.toml` so it's defined in a single place.


eniklas:
@derek-keeler if you know off-hand how to fix the CI failures, go for it. I'm donking. Will figure it out tomorrow if you don't.

src/gamatrix/__main__.py
```diff
@@ -445,9 +445,12 @@ def parse_cmdline(argv: List[str], docstr: str) -> Dict[str, Any]:
     )
     log = logging.getLogger()
 
+    version = pkg_resources.get_distribution("gamatrix").version
+
     opts = parse_cmdline(
         argv=sys.argv[1:],
         docstr=__doc__ if __doc__ is not None else "",
+        version=version,
```
eniklas, line 453: Without this new arg, `-v` was working fine but the tests failed with:

```
FAILED test/test_cmdline.py::test_new_cmdline_handling[No switches-commandline0-config_fields0-expected_values0] - NameError: name 'version' is not defined
FAILED test/test_cmdline.py::test_new_cmdline_handling[Assorted values all in one-commandline1-config_fields1-expected_values1] - NameError: name 'version' is not defined
FAILED test/test_cmdline.py::test_new_cmdline_handling[Only set the mode to server-commandline2-config_fields2-expected_values2] - NameError: name 'version' is not defined
FAILED test/test_cmdline.py::test_new_cmdline_handling[Allow the cache to update missing items.-commandline3-config_fields3-expected_values3] - NameError: name 'version' is not defined
```

Might be a better way to do it, I don't know docopt.

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
     - name: Analyze with mypy
       run: |
-        python -m mypy
+        python -m mypy --install-types --non-interactive
```
eniklas, line 39: mypy tests were failing with:

```
src/gamatrix/__main__.py:29: error: Library stubs not installed for "pkg_resources"  [import]
src/gamatrix/__main__.py:29: note: Hint: "python3 -m pip install types-setuptools"
src/gamatrix/__main__.py:29: note: (or run "mypy --install-types" to install all missing stub packages)
src/gamatrix/__main__.py:29: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

There doesn't seem to be a way to add command-line args to mypy in the toml file, so I had to add them here.

derek-keeler: This works just fine. I'll play around with this later and see if I can figure out how things are to be with `pyproject` and `mypy`. Also, I'll take a closer look at Ruff as it seems simpler. (There is also a VSCode plugin for it so local configuration should be a bit easier).

https://mypy.readthedocs.io/en/stable/config_file.html#example-pyproject-toml

src/gamatrix/__main__.py
derek-keeler: Yes, this is correct - but I don't understand why it was working before. Maybe defining `VERSION` is picked up by `docopt`... will look at it later.

However, for your change, I believe this is the _correct_ way to do things. Tell `docopt` what the version is explicitly.

derek-keeler (commented):
LGTM :ship:  it.

src/gamatrix/__main__.py
eniklas: Before, `VERSION` was coming in from an import, so it makes sense it was available before basically as a global var.

PR #122, opened 2023-02-12
Updated mypy dependencies (type stubs)

derek-keeler:
- also upgraded requests (minor)


derek-keeler:
@eniklas mypy fix was to just add a `type-` to the dev dependencies. Also updated requests to the latest (minor version update).

I tested on Python 11 and all seems well. Want to support that?

eniklas (approved):
> I tested on Python 11 and all seems well. Want to support that?

You mean Python 3.11? We already support it?

PR #123, opened 2023-02-13
add justfile and dev environment

eniklas:
This adds a [justfile](https://github.com/casey/just) (modern `make` replacement) with recipes to make development easier. Notably, it includes a `just dev` recipe that spins up the container and dumps you into it in editable mode, so you can dev directly in the container. I opted to move to the full `python:3.11` base as the `slim` base is too minimal for comfortable development (e.g., no `ps`). This bloats the container from 170MB to 965MB, but a 1GB container is still very reasonable so I think it's worth it.

The `justfile` only supports Linux for now. Just is cross-platform and recipes can be [OS aware](https://github.com/casey/just#recipe-attributes), so we can add Windows support in the future if desired. Another approach would be to run `just` from inside the container, but that will require installing it in the container along with configuring `git` and possibly `ssh`, which we can do if we like; I didn't want to go down that rabbit hole just now.


README.md
```diff
@@ -130,12 +132,10 @@ py -3 -m venv venv
 
 ```bash
 python -m pip install -U pip
-python -m pip install -e .[dev]
-python -m gamatrix
+python -m pip install .
```
eniklas, line 135: Since this is for normal running of the project, not dev, I don't think we want editable mode or to specify `.[dev]`.

Issue #120, opened 2023-02-12
Filter out expired Xbox game pass games


eniklas:
GOG Galaxy has an [issue](https://www.gog.com/forum/general_beta_gog_galaxy_2.0/suggestion_ability_to_remove_xbox_game_pass_games_from_owned) where, if you install a game pass game, it will be added to the "owned" view and will stay there even after the game leaves game pass. See if we can identify games in this situation and filter them out.

eniklas:
This is improved in https://github.com/eniklas/gamatrix/pull/159, but can't be fully fixed in gamatrix. I'm looking into whether it can be fixed in the integration.

eniklas:
The Xbox integration is not open source, so we can't improve it. The best workaround I've found is to disconnect the integration, which removes all Xbox titles from your library, and then reconnect it to get the currently available titles. We've done as much as we can do on the gamatrix side, closing.

PR #128, opened 2026-06-05
v2: rewrite to FastAPI/Lambda/DynamoDB on AWS

eniklas:
Replace the v1 Flask CLI/server with a FastAPI + Mangum app that runs on AWS Lambda (and locally in containers). DynamoDB replaces the SQLite/JSON cache, S3 + SQS drive async IGDB enrichment, and HTMX + Jinja2 render the UI. Adds custom JWT auth with SES password reset, issue #120 Game Pass filtering, and AWS CDK infrastructure. Domain/email and seed-user data are sourced from a private sibling repo so no PII lives here.


src/gamatrix/config.py
```diff
@@ -0,0 +1,108 @@
...
+    table_prefix: str = "gamatrix"
+
+    # --- S3 ---
+    upload_bucket: str = "gamatrix-gog-db-uploads"
+
+    # --- SQS (unset locally; the local_worker polls the jobs table instead) ---
+    enrichment_queue_url: str | None = None
+
+    # --- Auth ---
+    jwt_secret: str = "change-me-in-production"
```
derek-keeler: Is there anything forcing this to be set to something unique/unknown in production?

What would direct my attention to actually change this?

Can we put something in a deployment script that will override this to something - or better yet, is there someplace we can raise an exception to call it out (if it is found to be set to "change-me-in-production" still)?

src/gamatrix/auth/service.py
```diff
@@ -0,0 +1,157 @@
...
+
+    token = str(uuid.uuid4())
+    expires = datetime.now(timezone.utc) + timedelta(
+        minutes=settings.reset_token_ttl_minutes
+    )
+    repo.update_user(
+        email,
+        {"reset_token": token, "reset_token_expires": expires.isoformat()},
+    )
+    link = f"{settings.app_base_url}/auth/reset-password?token={token}&email={email}"
```
derek-keeler: Might want to URL-encode this string for the yutz's out there that use clever email tricks like `derek.keeler+gamatix-spam-catcher@derekscrud.ca` (or you will lose the `+` for example)... I'll put a test to add to catch this sort of thing into the test file for you...

tests/test_auth.py
```diff
@@ -0,0 +1,72 @@
...
+    assert service.complete_password_reset(repo, "user@x.com", token, "new")
+    assert service.authenticate(repo, "user@x.com", "new") is not None
+    # Token cleared after use.
+    assert repo.get_user("user@x.com").get("reset_token") is None
+
+
+def test_password_reset_unknown_email_is_silent(repo, settings, monkeypatch):
+    monkeypatch.setattr(service, "_send_email", lambda *a, **k: None)
+    # Should not raise for an unknown account.
+    service.begin_password_reset(repo, "ghost@x.com", settings)
```
derek-keeler, line 74:
```suggestion
    service.begin_password_reset(repo, "ghost@x.com", settings)
    
def test_password_reset_link_url_encodes_plus_email(repo, settings, monkeypatch):
     """Test that the reset link correctly URL-encodes email addresses with special characters."""
     sent = {}
     def fake_send(s, to, subject, body):
         sent["body"] = body
     monkeypatch.setattr(service, "_send_email", fake_send)
     repo.put_user(
         {
             "email": "user+games@x.com",
             "username": "User",
             "password_hash": service.hash_password("old"),
         }
     )
     service.begin_password_reset(repo, "user+games@x.com", settings)
     reset_link = next(line for line in sent["body"].splitlines() if "/auth/reset-password?" in line)
     query = parse_qs(urlparse(reset_link).query)
 
     assert query["email"] == ["user+games@x.com"]
```

derek-keeler: Stupid tricky spam catcher people.

derek-keeler: (NOTE: test written by GPT-5.4)

src/gamatrix/storage/dynamo.py
```diff
@@ -0,0 +1,240 @@
...
+            UpdateExpression=expr,
+            ExpressionAttributeNames=names,
+            ExpressionAttributeValues=values,
+        )
+
+    # ------------------------------------------------------------------
+    # user_libraries  (PK user_id, SK release_key)
+    # ------------------------------------------------------------------
+    def get_user_library(self, user_id: str) -> list[dict]:
+        resp = self._table(self.settings.libraries_table).query(
```
derek-keeler: Maybe pull all pages from these queries - our libraries are no joke! (Just like the query being done in `_scan` below...)

derek-keeler (commented):
Love it. Me and `gipity` came up with a few things you may want Klaus to look at further.

Can't wait to see the finished product!!

src/gamatrix/config.py
eniklas: Good catch — and it was actually worse than a config default: the CDK stack never set `JWT_SECRET` at all, so prod was signing sessions with this public value. Fixed in 87617c5:

- CDK now generates a stable `gamatrix/jwt-secret` in Secrets Manager (created once, survives redeploys), and the web Lambda reads it via `JWT_SECRET_NAME` (cached so warm invocations do not refetch per request).
- A `Settings` validator now raises on startup if the default is left set outside local dev, so an insecure deploy fails loudly instead of silently.

src/gamatrix/auth/service.py
eniklas: Fixed in 87617c5 — the address is now `urllib.parse.quote`d into the reset link, so `+` (and friends) survive the round-trip instead of decoding to a space.

tests/test_auth.py
eniklas: Added in 87617c5 — adapted as a standalone test with the `parse_qs`/`urlparse` imports wired in. Thanks (and thanks gipity 🙂).

src/gamatrix/storage/dynamo.py
eniklas: Fixed in 87617c5 — added a `_query_all` helper that follows `LastEvaluatedKey` (mirroring `_scan`), and routed both `get_user_library` and `get_owners_of_release` through it so large libraries are not truncated at the 1 MB query-page cap.

derek-keeler (approved):
🚢 it

Let's start adding features!

PR #132, opened 2026-06-06
Support multiple DNS names

eniklas:
This adds support for DNS aliases for the web app to prevent cert warnings. Only aliases in the same domain are supported for now.

### DIFF
--- a/infrastructure/cdk/config.py
+++ b/infrastructure/cdk/config.py
@@ -45,10 +45,22 @@ def load_deploy_config() -> DeployConfig:
     import yaml
 
     data = yaml.safe_load(path.read_text()) or {}
+
+    alias_domains = data.get("alias_domains") or []
+    # YAML happily parses `alias_domains: games.other.com` as a scalar string.
+    # Without this guard it would later be iterated character-by-character,
+    # silently wiring up bogus single-letter "domains". Require a real list.
+    if not isinstance(alias_domains, list):
+        kind = type(alias_domains).__name__
+        raise ValueError(f"alias_domains must be a list of hostnames, got {kind}")
+    alias_hosted_zone = data.get("alias_hosted_zone")
+    if alias_domains and not alias_hosted_zone:
+        raise ValueError("alias_hosted_zone is required when alias_domains is set")
+
     return DeployConfig(
         hosted_zone=data.get("hosted_zone"),
         site_domain=data.get("site_domain"),
         email_from=data.get("email_from"),
-        alias_hosted_zone=data.get("alias_hosted_zone"),
-        alias_domains=data.get("alias_domains") or [],
+        alias_hosted_zone=alias_hosted_zone,
+        alias_domains=alias_domains,
     )
--- a/infrastructure/cdk/config.py
+++ b/infrastructure/cdk/config.py
@@ -13,7 +13,7 @@ Gateway URL -- so anyone can clone and deploy this repo without owning a domain.
 from __future__ import annotations
 
 import os
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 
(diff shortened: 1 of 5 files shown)


infrastructure/cdk/config.py
```diff
@@ -44,4 +49,6 @@ def load_deploy_config() -> DeployConfig:
         hosted_zone=data.get("hosted_zone"),
         site_domain=data.get("site_domain"),
         email_from=data.get("email_from"),
+        alias_hosted_zone=data.get("alias_hosted_zone"),
+        alias_domains=data.get("alias_domains") or [],
```
derek-keeler: This is presumed to be a list. If the YAML defines it as a string it isn't being validated and it could result in some weird behaviour.

derek-keeler (commented):
One issue came up that I agree with (GPT 5.4 caught it)

infrastructure/cdk/config.py
derek-keeler: Test file to display the bad behaviour:

`tests/test_cdk_config.py`:

```python
"""Regression tests for CDK deploy-config alias validation.

These tests load the infrastructure deployment config from a temporary
``cdk-config.yaml`` file and assert that invalid alias-domain settings are
rejected before stack synthesis. The expected result is a fast failure for bad
input, rather than silently skipping alias wiring or treating a scalar string as
an iterable of one-character domains.
"""

from __future__ import annotations

import importlib.util
import sys
from pathlib import Path
from types import ModuleType

import pytest

PROJECT_ROOT = Path(__file__).resolve().parents[1]
CDK_CONFIG_PATH = PROJECT_ROOT / "infrastructure" / "cdk" / "config.py"


def _load_cdk_config_module() -> ModuleType:
    """Load the CDK config module directly from the infrastructure tree."""
    spec = importlib.util.spec_from_file_location("gamatrix_cdk_config_test", CDK_CONFIG_PATH)
    assert spec is not None
    assert spec.loader is not None

    module = importlib.util.module_from_spec(spec)
    sys.modules[spec.name] = module
    spec.loader.exec_module(module)
    return module


def test_load_deploy_config_rejects_alias_domains_scalar(tmp_path, monkeypatch):
    config_dir = tmp_path / "config"
    config_dir.mkdir()
    (config_dir / "cdk-config.yaml").write_text(
        "\n".join(
            [
                "hosted_zone: example.com",
                "site_domain: games.example.com",
                "alias_hosted_zone: other.com",
                "alias_domains: games.other.com",
            ]
        )
    )
    monkeypatch.setenv("GAMATRIX_CONFIG_DIR", str(config_dir))

    config = _load_cdk_config_module()

    with pytest.raises(ValueError, match="alias_domains"):
        config.load_deploy_config()


def test_load_deploy_config_requires_alias_hosted_zone(tmp_path, monkeypatch):
    config_dir = tmp_path / "config"
    config_dir.mkdir()
    (config_dir / "cdk-config.yaml").write_text(
        "\n".join(
            [
                "hosted_zone: example.com",
                "site_domain: games.example.com",
                "alias_domains:",
                "  - games.other.com",
            ]
        )
    )
    monkeypatch.setenv("GAMATRIX_CONFIG_DIR", str(config_dir))

    config = _load_cdk_config_module()

    with pytest.raises(ValueError, match="alias_hosted_zone"):
        config.load_deploy_config()
```

eniklas: Good catch — fixed in 10e2745. `load_deploy_config` now rejects a scalar `alias_domains` (raises `ValueError` instead of silently iterating the string character-by-character) and requires `alias_hosted_zone` whenever `alias_domains` is set. I also added your suggested `tests/test_cdk_config.py` (both cases pass). Thanks!

derek-keeler (approved):
Changes look fine. Ship it.

PR #135, opened 2026-06-07
Auto bump version on merge

eniklas:
This adds automatic versioning. Merge to `main` tags the new commit with the next SemVer (patch rev by default, major and minor bumps controlled by labels), and creates a new release. The version in the docker containers is handled dynamically at build time.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @derek-keeler @eniklas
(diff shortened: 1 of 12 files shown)


derek-keeler (approved):
Give'r

Looks fine. If this pops off and starts making you f-you money we can talk about locking down releases a bit further.

PR #136, opened 2026-06-07
Restore "Show product keys" option

eniklas:
The v1 UI had a **Show product keys** checkbox that added a Release Key column to the results table. It was dropped in the v2 rewrite — this brings it back.

## Changes
- `CompareOptions.show_keys`, parsed from the `show_keys` query flag (falling back to the saved preference) in `games/routes.py`
- `show_keys` added to `DEFAULT_PREFERENCES` and persisted by the preferences save route, so it can be saved as a default
- "Show product keys" checkbox on the filter bar (`games.html.jinja`) and the preferences page (`preferences.html.jinja`)
- Release Key column (`game.release_key`) in `games_table.html.jinja`, rendered only when enabled, in both list and grid views

## Notes
- In v2, cross-platform copies owned by the same people are merged into one row, so that row shows a single representative release key (matching v1's merge behavior).

## Testing
- `just check` is green (black, flake8, mypy, 32 tests)
- Render check confirms the column appears only when the option is enabled

### DIFF
--- a/src/gamatrix/games/preferences.py
+++ b/src/gamatrix/games/preferences.py
@@ -11,6 +11,7 @@ DEFAULT_PREFERENCES: dict[str, Any] = {
     "default_view": "list",
     "selected_users": "all",
     "exclusive": False,
+    "show_keys": False,
 }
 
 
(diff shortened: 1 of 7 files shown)


derek-keeler (approved):
Looks good - I assume this is a debugging QOL feature?

eniklas:
> Looks good - I assume this is a debugging QOL feature?

Yeah, it's really useful when debugging or if you just like "stats for nerds"

PR #137, opened 2026-06-07
Commit uv.lock and install from it in CI and Docker

eniklas:
## Should `uv.lock` be checked in? Yes.

gamatrix is an application (deployed to Lambda), not a redistributable library, so committing `uv.lock` is the recommended practice — it makes dependency resolution reproducible and gives a reviewable diff when deps change. The lock was simply never added (it's not in `.gitignore`).

But committing it alone only helps *local* `uv sync`. Previously **nothing else consumed the lock**:
- CI installed with `pip install .[dev,ci]`
- The Lambda image used `uv pip install --system .`

Both resolved fresh from `pyproject.toml`, so the lock could silently drift from what actually ships. This PR wires every build to the lock.

## Changes
- **Track `uv.lock`** (verified in sync via `uv lock --check`).
- **CI** (`.github/workflows/ci.yml`): switch to `uv`, install with `uv sync --frozen --extra dev --extra ci` (fails if the lock is stale), run checks via `uv run`. Adds `astral-sh/setup-uv` with caching.
- **Dockerfile** (dev + Lambda stages): export the frozen, hash-pinned lock to a requirements file and `uv pip install --system` it, then install the project with `--no-deps`. Keeps the `--system` install model the Lambda runtime needs, while pinning transitive deps with hashes.

## Notes
- Direct runtime deps in `pyproject.toml` are already `==`-pinned, so the lock's added value is pinning **transitive** deps with hashes.
- Validated locally: `uv export --frozen` produces a hashed requirements file (500 hash lines), `uv pip install -r` of it succeeds, and `uv pip install --no-deps .` builds `gamatrix` with the `SETUPTOOLS_SCM_PRETEND_VERSION` arg.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,31 +26,30 @@ jobs:
         # Full history + tags so setuptools_scm can derive the version.
         fetch-depth: 0
 
-    - name: Set up Python
-      uses: actions/setup-python@v3
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
       with:
-        python-version: ${{ matrix.python_version }}
+        enable-cache: true
+
+    - name: Set up Python
+      run: uv python install ${{ matrix.python_version }}
 
     - name: Install dependencies
-      run: |
-        python -m pip install -U pip
-        python -m pip install .[dev,ci]
+      # --frozen installs the exact versions in uv.lock and fails if the lock is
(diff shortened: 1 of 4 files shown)


derek-keeler (approved):
Yep fine.

<Soap Box>
Do note that `uv` is a bit frowned on in the Python community - is a for-profit company that hasn't been transparent about how they will support the community over the long term.
</Soap Box>

...meh

PR #138, opened 2026-06-07
Restore v1 grid ownership colors

eniklas:
In v1, the game **grid** view colored each per-user cell to show ownership: green when the user owned the game, red when they didn't. The v2 rewrite replaced this with a single `●` marker and no color. This brings the v1 scheme back.

## Changes
- `style.css`: add `--owned` / `--unowned` CSS vars using v1's exact colors (`rgb(182,215,168)` green / `rgb(234,153,153)` red) and `td.owned` / `td.unowned` rules.
- `games_table.html.jinja`: apply `owned`/`unowned` classes to grid cells. Since the cell color now conveys ownership, the redundant `●` marker is removed; the installed indicator (profile pic or ✓) is unchanged.

## Notes
- Only affects the grid view; the list view is untouched.
- Cell text uses the existing `--row-text` dark color, which stays readable on both green and red.

## Testing
- Render check: with two selected users, an owned+installed cell gets `class="owned"` and shows ✓, a non-owner cell gets `class="unowned"`, and `●` is gone.
- `just check` green (lint, mypy, 32 tests).

### DIFF
--- a/src/gamatrix/static/style.css
+++ b/src/gamatrix/static/style.css
@@ -6,6 +6,8 @@
   --row-a: rgb(180, 227, 239);
   --row-b: rgb(201, 235, 243);
   --row-text: #102027;
+  --owned: rgb(182, 215, 168);
+  --unowned: rgb(234, 153, 153);
 }
 
 body {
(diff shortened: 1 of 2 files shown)


derek-keeler (approved):
🚢 it

PR #140, opened 2026-06-07
Fix Include single-player toggle ignoring unchecked state

eniklas:
## Problem
"Include single-player" had no effect — games with max players = 1 showed whether or not the box was checked.

## Root cause
An unchecked HTML checkbox submits **no value**. In `_parse_options`, the flag resolver fell back to the saved preference whenever a param was absent:

```python
def flag(name, default):
    if name in qp:
        return qp[name] in ("true", "on", "1")
    return default   # absent -> saved pref, even when the user unchecked it
```

So it couldn't distinguish "user turned it off" from "not specified." Once a user saved `include_single_player=True`, unchecking the box fell back to the saved `True` and never took effect. Reproduced:

```
box unchecked, saved pref True -> include_single_player = True   (bug)
box checked                    -> include_single_player = True
```

This affected every boolean filter checkbox (single-player, installed-only, exclusive); single-player is just the most visible.

## Fix
Add a hidden `filters_active` marker to the `#filters` form. It's sent on every filter-form submission but absent on a bare page load, which lets the server tell the two apart:

```python
form_submitted = "filters_active" in qp
def flag(name, default):
    if name in qp:
        return qp[name] in ("true", "on", "1")
    return False if form_submitted else default
```

- **Form submitted** (HTMX table fetch / sort): absent checkbox → off.
- **Bare `/games` load**: no marker → saved prefs still apply, so the boxes render in their saved state.

## Tests
Added `tests/test_routes.py` covering: unchecked-overrides-saved-on, checked-turns-on, bare-load-uses-pref, and that the marker applies to all three boolean flags. `just check` green (lint, mypy, 36 tests).

## Note / out of scope
The list-valued filters have the same latent pattern (`qp.getlist("exclude") or prefs[...]`, and the user selection fallback): with all platform-exclude boxes unchecked you can't currently clear a saved exclude list. I kept this PR focused on the reported boolean-toggle bug; happy to extend the `filters_active` logic to those in a follow-up if you want.

### DIFF
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -47,10 +47,16 @@ def _parse_options(request: Request, user: dict) -> CompareOptions:
         else:
             selected = list(pref_users)
 
+    # When the filter form is submitted it includes a hidden `filters_active`
+    # marker. An unchecked checkbox sends no value, so without this marker we
+    # can't tell "user turned it off" from "not specified". When the form was
+    # submitted, an absent flag means off; on a bare page load, use the pref.
+    form_submitted = "filters_active" in qp
+
     def flag(name: str, default: bool) -> bool:
         if name in qp:
             return qp[name] in ("true", "on", "1")
-        return default
+        return False if form_submitted else default
 
     view = qp.get("view", prefs["default_view"])
     return CompareOptions(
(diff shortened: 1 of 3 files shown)


derek-keeler (approved):
Seems fine... Not an expert with this kind of flow but it seems logical.

PR #141, opened 2026-06-07
Apply filters_active marker to list-valued filters

eniklas:
Follow-up to #140. That PR fixed the **boolean** filter checkboxes; this extends the same `filters_active` marker to the **list-valued** filters, which had the identical latent bug.

## Problem
`_parse_options` resolved list filters as `qp.getlist(name) or prefs[name]`, so an empty submitted list fell back to the saved preference. Practical effects:
- Unchecking **all** platform-exclude boxes couldn't clear a saved exclude list.
- Deselecting **all** users silently reselected the saved set instead of showing nothing.

## Fix
Add a `multi()` helper mirroring `flag()` from #140:

```python
def multi(name, default):
    values = qp.getlist(name)
    if values or form_submitted:
        return values   # submitted form: empty means explicitly empty
    return default       # bare load: use saved pref
```

- `exclude_platforms` now uses `multi("exclude", prefs["exclude_platforms"])`.
- The user-selection fallback is gated on `not form_submitted`, so deselecting everyone yields an empty selection — and skips the `scan_users()` call that the fallback would otherwise make.

## Tests
Added cases to `tests/test_routes.py`: emptied exclude clears / partial exclude wins / bare-load uses saved exclude; deselect-all-users → empty on submit / bare-load uses saved users. `just check` green (lint, mypy, 41 tests).

## Note
Based on the #140 branch so the diff is just the follow-up. Once #140 merges, I'll retarget this to `master` (or it can be merged after #140).

### DIFF
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -36,9 +36,28 @@ def _parse_options(request: Request, user: dict) -> CompareOptions:
     prefs = merge_preferences(user.get("preferences", {}))
     qp = request.query_params
 
-    # User selection: explicit ?user=… params win; otherwise saved preference.
+    # When the filter form is submitted it includes a hidden `filters_active`
+    # marker. An off checkbox — or a list filter with everything deselected —
+    # sends no value, so without this marker we can't tell "user cleared it"
+    # from "not specified". When the form was submitted, an absent value means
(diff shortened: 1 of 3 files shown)


derek-keeler (approved):
Again, no expert here but the logic looks ok.

Issue #131, opened 2026-06-06
Progress bar displays inaccurate counts


eniklas:
The progress bar that indicates that enrichment is happening sometimes shows inaccurate counts (e.g., numerator larger than denominator):

`Updating game data from IGDB… 874 / 225`

eniklas:
Fixed in https://github.com/eniklas/gamatrix/pull/143.

PR #143, opened 2026-06-07
Fix stuck enrichment progress bar and inaccurate counts (#131)

eniklas:
## Problem

Two symptoms with one root cause: a job that never reaches a terminal status.

- **Stuck "Updating game data from IGDB…" bar.** It's not a cookie — the job id is re-derived on every `/games` load via `repo.get_active_job()`, which scans **all** jobs globally and returns any still `pending`/`running`. When the enricher Lambda crashes or hits its hard 15-min timeout, the `try/except` in `run_job` doesn't fire, so the job stays `running` forever. That one ghost job then re-attaches the progress bar on every page load, **for every user**, and blocks new enrichment from queueing.
- **Inaccurate counts (#131), e.g. `874 / 225`.** SQS is at-least-once. A job that runs longer than the 16-min visibility timeout (a big, rate-limited library) gets redelivered and re-runs concurrently. The progress write was an atomic `completed_count = completed_count + 1`, never reset, so overlapping runs pushed the count well past `total`.

## Fixes (app code)

- **Idempotent progress.** `increment_job_progress` → `set_job_progress(job_id, completed)` writes an **absolute** value from a locally-tracked counter, so redelivered/concurrent runs converge instead of inflating. It also stamps `updated_at`.
- **Staleness.** New `is_job_active` / `is_job_stale` helpers: an active job with no progress within `JOB_TIMEOUT_MINUTES` (20, above the 15-min enricher cap) is presumed dead. Measured from **last progress** (`updated_at`), not creation, so a slow-but-live job isn't reaped mid-run.
- `get_active_job()` skips stale jobs, so a ghost job can't pin the bar or block new enrichment.
- `job_status` route treats a stale job as `done`, so an already-displayed bar self-clears on its next 3s poll without a reload.
- Template clamps the rendered count to `total` as a display backstop.

## Tests

New `tests/test_jobs.py` (10 cases): active/stale classification, staleness measured from progress vs. creation, `get_active_job` skipping stale jobs, idempotent progress. Full suite **53 passed**; lint + typecheck clean.

## Follow-up

Infra backstop — DLQ + `maxReceiveCount` on `EnrichmentQueue` — will land in a separate PR. This PR does not actively flip stale jobs to `failed` in DynamoDB (they just age out of `get_active_job`); self-healing can ride along with the infra PR if wanted.

### DIFF
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -175,7 +175,9 @@ def job_status(
 ):
     job = repo.get_job(job_id)
     done = (
-        job is None or job["status"] in (JOB_COMPLETED, JOB_FAILED) or is_job_stale(job)
+        job is None
+        or job.get("status") in (JOB_COMPLETED, JOB_FAILED)
+        or is_job_stale(job)
     )
     return templates.TemplateResponse(
         request,
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -22,7 +22,7 @@ from gamatrix.games import service
 from gamatrix.games.preferences import merge_preferences
 from gamatrix.games.service import CompareOptions
 from gamatrix.helpers import parse_iso
-from gamatrix.jobs import create_enrichment_job
+from gamatrix.jobs import create_enrichment_job, is_job_stale
 from gamatrix.storage.dynamo import Repository
 from gamatrix.storage.queue import get_queue
 from gamatrix.templating import templates
@@ -174,7 +174,9 @@ def job_status(
     repo: Repository = Depends(get_repo),
 ):
     job = repo.get_job(job_id)
-    done = job is None or job["status"] in (JOB_COMPLETED, JOB_FAILED)
+    done = (
+        job is None or job["status"] in (JOB_COMPLETED, JOB_FAILED) or is_job_stale(job)
+    )
     return templates.TemplateResponse(
         request,
         "job_status.html.jinja",
--- a/src/gamatrix/igdb/enricher.py
+++ b/src/gamatrix/igdb/enricher.py
@@ -34,7 +34,7 @@ async def run_job(
         return
 
     repo.update_job(job_id, {"status": JOB_RUNNING, "updated_at": now_iso()})
-    release_keys: list[str] = job["release_keys"]
+    release_keys: list[str] = job.get("release_keys", [])
 
     # Group release keys by the IGDB key so games shared across platforms
     # (e.g. a Steam and a GOG copy) only cost one set of API calls.
--- a/src/gamatrix/igdb/enricher.py
+++ b/src/gamatrix/igdb/enricher.py
@@ -33,7 +33,7 @@ async def run_job(
         log.error("Enrichment job %s not found", job_id)
         return
 
-    repo.update_job(job_id, {"status": JOB_RUNNING})
+    repo.update_job(job_id, {"status": JOB_RUNNING, "updated_at": now_iso()})
     release_keys: list[str] = job["release_keys"]
 
     # Group release keys by the IGDB key so games shared across platforms
@@ -49,6 +49,7 @@ async def run_job(
         by_igdb_key.setdefault(game["igdb_key"], []).append(rk)
 
     client_id, client_secret = resolve_igdb_credentials(settings)
+    completed = 0
     try:
         async with IGDBClient(client_id, client_secret) as client:
             for igdb_key, rks in by_igdb_key.items():
@@ -61,7 +62,11 @@ async def run_job(
                     meta = GameMetadata()
                 for rk in rks:
                     _write_metadata(repo, games[rk], meta)
-                    repo.increment_job_progress(job_id, 1)
+                    completed += 1
+                    # Absolute, not incremental: a redelivered or concurrent
+                    # run converges here instead of pushing the count past
+                    # `total` (see #131).
+                    repo.set_job_progress(job_id, completed)
     except Exception:
         log.exception("Enrichment job %s failed", job_id)
         repo.update_job(job_id, {"status": JOB_FAILED, "completed_at": now_iso()})
--- a/src/gamatrix/jobs.py
+++ b/src/gamatrix/jobs.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import uuid
 from datetime import datetime, timedelta, timezone
+from typing import NotRequired, TypedDict
 
 from gamatrix.constants import (
     JOB_PENDING,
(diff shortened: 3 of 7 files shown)


src/gamatrix/storage/dynamo.py
```diff
@@ -195,13 +199,17 @@ def list_pending_jobs(self) -> list[dict]:
...
 
     def get_active_job(self) -> dict | None:
-        """Return the most recently created pending-or-running job, if any."""
-        from gamatrix.constants import JOB_PENDING, JOB_RUNNING
+        """Return the most recently created pending-or-running job, if any.
+
+        Stale jobs (presumed-dead enrichers, see `is_job_stale`) are skipped so
+        a job that never reached a terminal status can't pin the progress bar
+        on every page load, nor block new enrichment from being queued."""
+        from gamatrix.jobs import is_job_active, is_job_stale
```
derek-keeler, line 213: Why is this import local?

src/gamatrix/jobs.py
```diff
@@ -4,15 +4,42 @@
...
+from gamatrix.helpers import now_iso, parse_iso
 from gamatrix.storage.dynamo import Repository
 from gamatrix.storage.queue import EnrichmentQueue
 
 log = logging.getLogger(__name__)
 
 
+def is_job_active(job: dict) -> bool:
+    """True while a job is still pending or running (not yet terminal)."""
+    return job.get("status") in (JOB_PENDING, JOB_RUNNING)
```
derek-keeler, line 42: Might want to write a log message here to call out when this has been called with an incorrect 'job' dict. But this might be a bit pedantic... (`job` is just a dict so anything could be sent in and that might become confounding when we add future features here.)

```suggestion
    if "status" not in job:
        log.debug("Incorrect `job` object sent to 'is_job_active'")
    return job.get("status") in (JOB_PENDING, JOB_RUNNING)
```

src/gamatrix/jobs.py
derek-keeler: Consider making this a type? Might help with the comment I made about checking for fields in debug mode.

tests/test_jobs.py
```diff
@@ -0,0 +1,108 @@
...
+def _job(repo, **overrides):
+    job = {
+        "job_id": overrides.get("job_id", "j1"),
+        "status": JOB_RUNNING,
+        "created_at": now_iso(),
+        "completed_at": None,
+        "release_keys": ["steam_1", "steam_2"],
+        "total": 2,
+        "completed_count": 0,
+    }
```
derek-keeler, line 35: Another use of this 'job' dict - yeah, a type might be useful here. (datatype would be fine)

src/gamatrix/games/routes.py
```diff
@@ -174,7 +174,9 @@ def job_status(
     repo: Repository = Depends(get_repo),
 ):
     job = repo.get_job(job_id)
-    done = job is None or job["status"] in (JOB_COMPLETED, JOB_FAILED)
+    done = (
+        job is None or job["status"] in (JOB_COMPLETED, JOB_FAILED) or is_job_stale(job)
```
derek-keeler: Might want to use `job.get("status") instead?


```suggestion
        job is None or job.get("status", None) in (JOB_COMPLETED, JOB_FAILED) or is_job_stale(job)
```

src/gamatrix/igdb/enricher.py
```diff
@@ -33,7 +33,7 @@ async def run_job(
         log.error("Enrichment job %s not found", job_id)
         return
 
-    repo.update_job(job_id, {"status": JOB_RUNNING})
+    repo.update_job(job_id, {"status": JOB_RUNNING, "updated_at": now_iso()})
     release_keys: list[str] = job["release_keys"]
```
derek-keeler: May want to protect against missing data...

```suggestion
    release_keys: list[str] = job.get("release_keys", [])
```

derek-keeler (commented):
Have a look at my suggestions - this `job` dict is getting passed around a bunch with assumptions about its content. Might want to make it an object or a datatype? (Feel free to say nah and forge ahead - ducktyping for the win)

src/gamatrix/storage/dynamo.py
eniklas: Circular import: \`gamatrix.jobs\` imports \`Repository\` from this module, so a top-level \`from gamatrix.jobs import ...\` here would cycle. Added an inline comment saying so in 2f8b972. (The new \`JobRecord\` type is imported under \`TYPE_CHECKING\` for the same reason — annotation-only, no runtime import.)

src/gamatrix/jobs.py
eniklas: Went with the type instead of a debug log: \`is_job_active\` now takes \`JobRecord\`, so a wrong-shaped argument is a mypy error rather than something we discover at runtime. The existing \`job.get("status")\` already returns \`None\` → \`False\` for a record missing the field, so it degrades safely without adding log noise on every poll. (2f8b972)

src/gamatrix/jobs.py
eniklas: Done — added a \`JobRecord\` TypedDict in 2f8b972 and threaded it through the helpers and the \`put_job\`/\`get_job\`/\`get_active_job\` repository methods. This dict is now annotated as \`JobRecord\`, so a missing/renamed field is caught by mypy.

tests/test_jobs.py
eniklas: The type (\`JobRecord\`) now exists as of 2f8b972. I deliberately left this test helper on a plain dict, though: several cases pass intentionally-partial records (e.g. \`{"status": JOB_PENDING}\`) to exercise the missing-field paths in \`is_job_stale\`, and mypy is scoped to \`files = "src"\` so tests aren't typechecked regardless. Tightening the helper would fight those edge-case inputs for no checking benefit.

src/gamatrix/games/routes.py
eniklas: Done in 2f8b972 — \`job_status\` now uses \`job.get("status")\`.

src/gamatrix/igdb/enricher.py
eniklas: Applied in 2f8b972 — \`run_job\` now reads \`job.get("release_keys", [])\`.

eniklas:
Thanks for the review @derek-keeler. Addressed in 2f8b972.

On the general "make `job` a type" note: agreed, went with a `JobRecord` TypedDict in `jobs.py` rather than a dataclass, since the record crosses the DynamoDB boundary as a plain dict (the repo's `_from_dynamo` returns `Any`) — a TypedDict gives mypy-checked field access without a serialize/deserialize layer or any runtime behavior change. It's threaded through the helpers and the `put_job`/`get_job`/`get_active_job` methods; the storage layer casts the untyped scan to `list[JobRecord]` at the boundary. I also hardened the two raw subscripts you flagged (`job.get("status")`, `job.get("release_keys", [])`) so a partial record degrades instead of raising.

Full suite still 53 passed; lint + typecheck clean.

tests/test_jobs.py
derek-keeler: I blame my youthful exuberance.

derek-keeler (approved):
Thanks for the changes, looking good. 🚢 it

PR #144, opened 2026-06-07
Add DLQ for enrichment queue and self-heal stale jobs

eniklas:
Infra backstop to #143 (stuck progress bar / inaccurate counts). **Stacked on `fix-stuck-progress-bar`** — review/merge that first; this PR's base will retarget to `master` once it lands.

## What

- **Dead-letter queue.** `EnrichmentQueue` gets a DLQ with `max_receive_count=3`. Previously a failing or timed-out job was redelivered for the full 1-day retention, and *each* redelivery re-ran the job — the engine behind #131's runaway counts. Capping redeliveries stops the storm (and pairs with #143's absolute progress write so the count can't inflate even within those 3 attempts). Dead messages are kept 14 days for inspection / manual redrive.
- **Self-healing.** `Repository.fail_stale_jobs()` marks presumed-dead jobs (`is_job_stale` from #143) as `failed`. `get_active_job()` now reaps them in passing during its existing scan, so a crashed or hard-timed-out enricher's job record is corrected instead of sitting `running` forever. The DLQ only caps redeliveries — it can't update the DynamoDB job record — so this app-side sweep is the necessary complement.

## Why a sweep rather than a DLQ-consumer Lambda

Marking the job `failed` event-driven from the DLQ would need a new Lambda + wiring for marginal benefit: the sweep runs on the `/games` load that's about to read the job anyway, which is exactly when a stale record matters. Healing is therefore lazy (on next `/games` load) but precisely timed. If we later want eager cleanup, a scheduled sweep or DLQ consumer can call the same `fail_stale_jobs()`.

## Tests

`tests/test_jobs.py` gains coverage for `fail_stale_jobs` (reaps only stale *active* jobs; leaves fresh/terminal ones) and `get_active_job` marking a stale job `failed`. Full suite **55 passed**; lint + typecheck clean.

The CDK DLQ change isn't unit-tested — the existing CDK tests only cover config validation, and synthesis here needs Docker image bundling. Verified by inspection.

### DIFF
--- a/src/gamatrix/storage/dynamo.py
+++ b/src/gamatrix/storage/dynamo.py
@@ -8,13 +8,17 @@ dynamodb-local in development; only the endpoint URL differs.
 from __future__ import annotations
 
 import decimal
-from typing import Any, Iterable
+from typing import TYPE_CHECKING, Any, Iterable, cast
 
 import boto3
 from boto3.dynamodb.conditions import Key
 
 from gamatrix.config import Settings, get_settings
 from gamatrix.helpers import now_iso
+
+if TYPE_CHECKING:
+    # Annotation-only; importing at runtime would cycle (jobs imports Repository).
+    from gamatrix.jobs import JobRecord
 
 
 def _to_dynamo(value: Any) -> Any:
@@ -160,10 +164,10 @@ class Repository:
     # ------------------------------------------------------------------
     # enrichment_jobs
     # ------------------------------------------------------------------
-    def put_job(self, job: dict) -> None:
+    def put_job(self, job: JobRecord) -> None:
         self._table(self.settings.jobs_table).put_item(Item=_to_dynamo(job))
 
-    def get_job(self, job_id: str) -> dict | None:
+    def get_job(self, job_id: str) -> JobRecord | None:
         resp = self._table(self.settings.jobs_table).get_item(Key={"job_id": job_id})
         item = resp.get("Item")
         return _from_dynamo(item) if item else None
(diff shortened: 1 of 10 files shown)


src/gamatrix/storage/dynamo.py
```diff
@@ -198,19 +198,42 @@ def list_pending_jobs(self) -> list[dict]:
...
+        from gamatrix.constants import JOB_FAILED
+        from gamatrix.jobs import is_job_active, is_job_stale
+
+        if jobs is None:
+            jobs = self._scan(self.settings.jobs_table)
+        reaped: list[str] = []
+        for j in jobs:
+            if is_job_active(j) and is_job_stale(j):
+                self.update_job(
+                    j["job_id"],
```
derek-keeler, line 225: More to do with job-as-type (see comment in other PR). Might want to switch to using `.get(<val>, <def>)` patters instead. Better yet, switch to using a common type for `job`...

eniklas: Done — went with the "common type for \`job\`" option. After rebasing onto the updated #143, this branch picks up the \`JobRecord\` TypedDict, and the DLQ code here now uses it: \`fail_stale_jobs(jobs: Iterable[JobRecord] | None = ...)\` and \`get_active_job() -> JobRecord | None\` (9b11729). So the \`j["job_id"]\` access is now a mypy-checked required field rather than an unguarded subscript on an untyped dict. The untyped DynamoDB scan is cast to \`list[JobRecord]\` at the boundary.

Worth noting: #144 targets #143, so this only goes green once #143 merges — the type itself lives there.

derek-keeler (approved):
ship this with #143

PR #145, opened 2026-06-07
Make agent instructions work for non-Claude agent frameworks

derek-keeler:
## What

Add support for Copilot (and Codex... etc...) without taking anything away from Claude Code.

## How

- CLAUDE.md becomes a thin wrapper pointing to the generic AGENTS.md
- .github/copilot-instructions is a thin wrapper also pointing to AGENTS.md
- Everything from CLAUDE.md moved into AGENTS.md, with some restructuring

NOTE: It is unclear if the restructuring is necessary, Copilot _did_ make it slightly more straightforward though.

## Checklist

- [ ] PR checks pass
- [x] ~~Tests added/updated for the change~~ N/A
- [x] ~~Version label applied if this is more than a patch~~ just a patch.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,4 +1,4 @@
-# CLAUDE.md
+# Copilot Instructions for Gamatrix
 
 The canonical shared agent guidance for this repository lives in `AGENTS.md`.
 
@@ -8,4 +8,4 @@ Start by reading `AGENTS.md` and treat it as the source of truth for:
 - high-level architecture and data flow
 - repository-specific conventions and patterns
 
-Keep this file thin. Add Claude-specific notes here only when they cannot live in the shared `AGENTS.md`.
+Keep this file thin. Add Copilot-specific notes here only when they cannot live in the shared `AGENTS.md`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,6 +2,10 @@
 
 This file is the canonical repository guidance for AI coding agents in this repo. Keep shared guidance here, and keep `CLAUDE.md` and `.github/copilot-instructions.md` as thin wrappers that point back to this file.
 
+## Project Overview
+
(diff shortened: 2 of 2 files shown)


eniklas (commented):
Reviewed as a docs refactor — mechanics are right and every factual claim in the new `AGENTS.md` checks out against the codebase (`games/service.py`, `templating.py`, `games/preferences.py`, the `get_repo()`/`get_s3()`/`get_queue()`/`current_user_api()`/`merge_preferences()`/`_parse_options()` references, the `(slug, owners)` merge at `service.py:147`, enricher grouping by `igdb_key`, and the `-k merge_cross_platform_duplicates` test all verified). The fan-out pattern (canonical `AGENTS.md` + thin `CLAUDE.md` and `.github/copilot-instructions.md` wrappers) and file paths are correct.

A few non-blocking notes, roughly in priority order:

1. **Lost the "what is this product" intro (content regression).** The old `CLAUDE.md` opened with a Project Overview paragraph (compares game libraries across users → upload GOG Galaxy SQLite → enrich via IGDB). `AGENTS.md` jumps straight from build commands into architecture, so a cold agent loses the high-level framing. Worth re-adding 2–3 sentences at the top.

2. **Architecture is described twice, in two styles.** The "Architecture" section gives a prose data-flow, then "Key conventions", then the doc snaps back to the old bullet blocks (Routers / Background Lambdas / Storage / IGDB client / Config / Templates / Tests). It reads disjointed — the component inventory is split across two places with conventions wedged between. This is the "unclear if restructuring is necessary" point from the description; I'd either fold the bullet inventory next to the prose or move conventions to the end.

3. **Two `preferences.py` files — minor ambiguity.** The Routers bullet says `preferences.py` (the top-level `src/gamatrix/preferences.py`, registered in `app.py`) while conventions reference `merge_preferences()` in `games/preferences.py`. Both are accurate, but a path qualifier on the router bullet would prevent conflation.

4. **Nit:** a couple of double blank lines (after the Architecture intro, after the versioning bullet).

Items 1 and 2 are the ones I'd address before merging; 3 and 4 are polish.

derek-keeler:
1. Copied the original Project Overview back to the top of the file - I agree it was odd to move it.
2. Moved "Key conventions" to the end after the architecture discussion. Better for human eyeballs.
3. Added explicit links to the specific files each instance was describing.

eniklas (approved):

PR #147, opened 2026-06-08
Use IGDB-style slugs as the internal slug format

eniklas:
## Why

The internal slug was lowercase-alphanumeric with separators **deleted**, so `"Ground Branch"` → `"groundbranch"`. That same value was used to query IGDB by slug, but IGDB slugs are **hyphenated** (`ground-branch`), so the `where slug = "..."` lookup never matched and titles fell through to the fuzzy fallbacks (sometimes landing at `not_found`, e.g. Ground Branch).

The old format only existed to stay compatible with v1 config/metadata slugs during import. That import is done, so we can switch to a single, better format.

## What

`get_slug_from_title` now produces a standard slugify and is used everywhere (game storage, override matching, override import, and the IGDB slug lookup):

- lowercase
- transliterate accents (`Pokémon` → `pokemon`)
- drop apostrophes entirely (`Avallac'h` → `avallach`, not `avallac-h`)
- collapse every other run of non-alphanumerics into a single `-`, trimmed

The old alphanumeric-only function is removed (one name, nothing to keep in sync).

## Data migration

Switching the format means already-stored values need re-keying. This has been done already with `migrate_metadata.py` (which I'm including as it could be useful to other users) and another script which is not included as it's only useful for migration from a previous v2 version.

- `migrate_metadata.py` gains `--wipe` (backed by `Repository.clear_metadata`) to clear stale old-format override keys before re-importing from the v1 config. Old slugs are lossy, so re-import is the only way to get new keys.

## Tests

Added `tests/test_helpers.py` covering the slugify cases above. `just check` (black + flake8 + mypy + pytest) passes; mypy now also covers `scripts/`.


tests/test_helpers.py
```diff
@@ -0,0 +1,24 @@
...
+
+@pytest.mark.parametrize(
+    "title,expected",
+    [
+        ("Ground Branch", "ground-branch"),
+        ("BioShock Infinite", "bioshock-infinite"),
+        ("Alan Wake 2", "alan-wake-2"),
+        # Apostrophes are dropped, not turned into separators.
+        ("Avallac'h", "avallach"),
+        ("Tom Clancy's Rainbow Six® Siege", "tom-clancys-rainbow-six-siege"),
```
derek-keeler, line 16: _oh ubisoft_ 🤦

derek-keeler (approved):
Good stuff - no more waiting for IGDB!

Would a backup of metadata for each subsequent upgrade (infra-only task) make sense? (Not this PR - that's a different issue altogether - automatic upgrade deployments should back everything up, maybe?)

eniklas:
> Good stuff - no more waiting for IGDB!

To be clear, there is no performance improvement here, just better matching in IGDB and getting rid of a separate slug we were using internally.

> Would a backup of metadata for each subsequent upgrade (infra-only task) make sense? (Not this PR - that's a different issue altogether - automatic upgrade deployments should back everything up, maybe?)

Can't think of a use for backups off-hand. The Dynamo data persists across deployments. We could consider this if there was a breaking change (e.g. major Dynamo upgrade that was destructive) but even then we'd probably just repopulate the data from IGDB.

PR #148, opened 2026-06-08
Fix duplicate DB upload ingest

derek-keeler:
## What

While running gamatrix (v2) locally I tripped on a couple of gotchas. (Note that my setup is Windows and this change may be "just the wrong setup" kind of problem... please do look closely!).

- This was a real production bug: both the AWS parser Lambda (lambdas/db_parser/handler.py) and local /upload/complete use the same ingest_db_file() path.
- There were two coupled issues:
    1. GogDBParser.parse() could emit duplicate library rows for the same release_key when the DB contained duplicate allGameReleases rows.
    2. Repository.replace_user_library() deleted and rewrote the same keys in one batch write, so re-uploading the same DB could wipe out the library or behave incorrectly.
- Local debugging has a network routing issue

1. Resolution of names: Gamatrix app was generating pre-signed upload URL with Docker’s internal hostname `minio:9000`, which the browser couldn't resolve from the host machine. That meant the upload stalled or failed before /upload/complete ever ran.

1. Post-upload ingestion duplication bug: Error was `botocore.exceptions.ClientError: ... BatchWriteItem operation: Provided list of item keys contains duplicates`. (src: src\gamatrix\storage\dynamo.py:148). Uploading my GOG DB twice produced duplicate library rows for the same DynamoDB key (user_id, release_key), and DynamoDB batch writes reject duplicate keys in a single batch. The upload of my GOG DB finished, but the ingestion failed while replacing the user library. Production bug: the AWS parser Lambda and local /upload/complete both call the same ingest_db_file(). 

## How

I’m adding two regressions: one for the parser emitting duplicate release keys from realistic GOG rows, and one for repeated ingest of the same DB staying idempotent. Fix the parser so production and local uploads both use the deduped path. Fixed the parser at the source so duplicate platform-list rows collapse to one library entry per release key before ingest ever touches DynamoDB. 

Also: I found that the new local .env was bleeding into the test config, so pytest started talking to the live local DynamoDB instead of moto. Isolated the tests from local endpoints.

`replace_user_library()` deletes and rewrites the same keys in one batch, which can wipe out the library on re-upload. Fixed that storage path so repeated uploads of the same data remains consistent (no changes applied unless changes necessary). The storage layer now treats repeated uploads as upserts instead of delete-then-reinsert for the same keys.

### Summary

- src/gamatrix/gogdb/parser.py - Dedupes parsed library entries by release_key while preserving install state.
- src/gamatrix/storage/dynamo.py - replace_user_library() now behaves like an upsert for incoming keys and only deletes stale keys, making repeated uploads of the same DB safe.
- Added regressions in tests/test_gogdb.py for: - duplicate parser rows from realistic GOG data
- repeated ingest of the same DB remaining idempotent
- Added tests/test_s3.py earlier for the local presigned-upload URL fix.
- Updated tests/conftest.py so pytest stays isolated from the local .env endpoints.

## Checklist

- [ ] PR checks pass


lambdas/__init__.py
```diff
@@ -0,0 +1 @@
+
```
eniklas: This file looks stray. Nothing imports `lambdas` as a package, `setuptools.packages.find` only scans `src`, and the actual handler dirs (`db_parser/`, `igdb_enricher/`) have no `__init__.py`. It has no effect on this fix — likely a leftover from debugging. Suggest dropping it unless there's a specific reason (making the parent a package while the subdirs aren't is a bit confusing).

src/gamatrix/storage/dynamo.py
```diff
@@ -135,15 +135,31 @@ def replace_user_library(self, user_id: str, entries: list[dict]) -> None:
...
         table = self._table(self.settings.libraries_table)
         existing = self.get_user_library(user_id)
+        incoming: dict[str, dict] = {}
+        for entry in entries:
+            release_key = entry["release_key"]
+            current = incoming.get(release_key)
+            if current is None:
+                incoming[release_key] = {**entry}
+                continue
+            current["installed"] = current.get("installed", False) or entry.get(
```
eniklas, line 145: Non-blocking: this dedup now overlaps with the parser fix, which already collapses duplicate `release_key`s. Keeping it is defensible belt-and-suspenders since `replace_user_library` is a general repo API and shouldn't trust callers — in fact this dynamo-side dedup alone is sufficient to fix the batch error. Just noting that, given the parser dedup, the non-`installed` last-non-None merge branch is effectively dead in this flow (all entries for a key are identical). Harmless.

eniklas (commented):
**Review: Approve.** Two real, coupled production bugs plus a local-dev upload fix, backed by regression tests I verified actually reproduce the failures.

I reverted the source changes to `master` while keeping the new tests — both fail with exactly `botocore...ValidationException ... BatchWriteItem ... Provided list of item keys contains duplicates`, and pass with the fix. All checks pass locally (47 tests, `black`, `flake8`, `mypy`).

**What's good**
- **Parser** (`parser.py`): `parsed.entries` got an append per occurrence while the `seen` set only deduped `parsed.games`; duplicate platform-list rows → duplicate `release_key` entries → duplicate `put_item` in one batch. Now keyed by `release_key` with `installed` OR-merged. Correct.
- **Storage** (`dynamo.py`): the old delete-all + put-all in a single `batch_writer` put the same key in both a delete and a put within one `BatchWriteItem` on re-upload → rejected. Now only stale keys (`existing − incoming`) are deleted and the rest upserted. Correct and idempotent.
- **S3 public endpoint**: for a presigned *POST* the signature lives in the form `fields`, not the URL, so rewriting only scheme+netloc (preserving path/query) is safe and keeps MinIO path-style `/bucket`. Properly gated and documented in `.env-sample`.
- **Test isolation** (`conftest.py`): pinning endpoint URLs to `None` stops `.env` leaking the live local DynamoDB into moto-backed tests.

Minor, non-blocking points inline below.

src/gamatrix/storage/s3.py
```diff
@@ -9,6 +9,7 @@
 from __future__ import annotations
 
 import boto3
+from urllib.parse import urlsplit, urlunsplit
```
eniklas, line 11: Trivial nit: stdlib import placed after the third-party `import boto3`. `black`+`flake8` don't care (no `isort` here), so feel free to ignore.

lambdas/__init__.py
eniklas: This is the only one I'd like done before approving, one less file.

derek-keeler: I found this when I ran `mypy .` locally - however it doesn't seem to be affecting the PR runner. I will lose the file since it doesn't have anything to do at all with my change. It is weird that running mypy locally shows errors - I'm not using `uv` locally so there's some magic voodoo going on.

derek-keeler: Done!

src/gamatrix/storage/dynamo.py
derek-keeler: Noted.

src/gamatrix/storage/s3.py
derek-keeler: 🤦🤦🤦

Bad Python-dev no cookie.

derek-keeler: Fixed.

derek-keeler: My honour has been restored.

eniklas (approved):
:shipit:

Issue #133, opened 2026-06-06
Allow users to update their profile pic and display name


eniklas:
Add an option for the user to upload their own profile pic and set their display name.

eniklas:
Done in https://github.com/eniklas/gamatrix/pull/149 and https://github.com/eniklas/gamatrix/pull/150.

PR #149, opened 2026-06-08
Allow users to edit their display name

eniklas:
Part of #133 (display-name half; profile-pic upload deferred — see below).

## What

Adds a **Display name** form to the preferences page so a signed-in user can change their `username`.

- New `POST /profile` endpoint validates and normalizes the input — trims/collapses whitespace, rejects empty, caps at `DISPLAY_NAME_MAX_LENGTH` (32) — then persists via `update_user`.
- `username` is read live everywhere through `scan_users()`, so the new name shows up across the games/preferences views immediately with no other changes.
- HTMX form shows inline "Saved ✓" / validation error without a page reload.

## Tests

`tests/test_preferences.py` covers the validation helper (`clean_display_name`) and the persisted update. `just check` (black + flake8 + mypy + pytest) passes.

## Why profile pics are deferred (the architectural note from #133)

Today profile pics are static `.png` files committed to the repo and served from `/static/profile_img/` (`games_table.html.jinja:25`). On Lambda that directory is **baked into the read-only container image**, so user-uploaded pics can't be written there. Supporting uploads requires:

- storing pics in S3 (the existing upload bucket under a non-`uploads/` prefix, so they don't trip the parser-Lambda S3 trigger filtered to `prefix="uploads/"` in `gamatrix_stack.py`), and
- a serving path — a `/profile_img/{user_id}` web route that streams from S3 with cache headers (the agreed approach), plus a through-Lambda multipart upload that validates/resizes.

That's a meaningfully larger change, so it's tracked as a follow-up rather than bundled here. Display-name editing is the trivial, self-contained half and ships first.

### DIFF
--- a/src/gamatrix/constants.py
+++ b/src/gamatrix/constants.py
@@ -4,6 +4,9 @@
 UPLOAD_ALLOWED_EXTENSIONS = ["db"]
 UPLOAD_MAX_SIZE = 300 * 1024 * 1024
 
+# Bounds for a user-chosen display name (the `username` field).
+DISPLAY_NAME_MAX_LENGTH = 32
+
 # Full mapping is at https://api-docs.igdb.com/#external-game-enums, but only
 # Steam and GOG actually work; the other platforms' IDs don't match IGDB.
 IGDB_PLATFORM_ID = {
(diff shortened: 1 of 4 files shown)


derek-keeler (approved):
Seems ok - not well versed in sanitizing web form details in Python, but this seems correct to me.

PR #150, opened 2026-06-08
Allow users to upload a profile picture

eniklas:
Completes #133. **Stacked on #149** (display name) — base is `user-display-name`, so this diff shows only the profile-pic changes. Merge #149 first, or rebase this onto master after it lands.

## What

Adds a **Profile picture** uploader to the preferences page (with a live preview), so users can replace the seeded static avatars with their own image.

## Architecture

- **Storage:** uploaded pics are stored in a dedicated `profile_pics` DynamoDB table keyed by `user_id` (processed 128px PNGs, ~tens of KB, comfortably under the 400KB item limit). Serving is a single keyed `get_item` — no full-table scan. Adds one table (CDK + `init_local` + a `Settings` property); no S3 involvement for this feature.
- **Upload path:** the image is sent as the **raw request body** (not multipart) and read from `request.stream()` with an early size cap, so an oversized body is aborted before it buffers to memory/disk. It's then processed with Pillow — decoded (validates it's really an image), bounds-checked against a max pixel count *before* full decode (decompression-bomb guard), resized to a ≤128px PNG, and re-encoded (strips EXIF).
- **Serving:** `GET /profile_img/{user_id}` streams the pic from DynamoDB, gated to logged-in users, with `Cache-Control: private` (the route is auth-gated). URLs carry a `?v=` cache-buster derived from the last update, so the browser cache stays fresh.
- **Resolution:** a `pic_url(user)` helper (registered as a Jinja global) returns the uploaded pic via the route (keyed off `pic_updated`), else the seeded static file, else `None`. The games/preferences templates use it instead of hardcoding `/static/profile_img/...`.

New user field: `pic_updated` (epoch; doubles as the cache-buster and the "has an upload" flag). Adds one dependency: `pillow`.

## Tests

- `tests/test_images.py` — resize/aspect/format, non-image rejection, decompression-bomb rejection.
- `tests/test_preferences.py` — `pic_url` resolution, DynamoDB pic round-trip, missing-pic None.
- Verified end-to-end against moto: pre-upload 404 → non-image 400 → oversized 400 (early abort) → upload 200 (cache-busted URL) → serve returns a 128px `image/png` with `Cache-Control: private`.

`just check` (black + flake8 + mypy + pytest, 56 tests) passes.

### DIFF
--- a/src/gamatrix/images.py
+++ b/src/gamatrix/images.py
@@ -1,9 +1,14 @@
 """Profile-picture image processing.
 
-User-uploaded images are normalized server-side before they reach S3: decoded
+User-uploaded images are normalized server-side before they are stored: decoded
 (which validates that the bytes really are an image), flattened onto a
 transparent canvas, downscaled to a small square-ish thumbnail, and re-encoded
 as PNG. Re-encoding also strips any EXIF/metadata the original carried.
+
+Decoding is the dangerous step. A small file can describe an enormous bitmap (a
+"decompression bomb"); fully decoding it can exhaust a worker's CPU/RAM. We
+guard against that twice: the header's declared dimensions are checked before
+any pixels are read, and Pillow's own bomb detector is left armed as a backstop.
 """
 
 from __future__ import annotations
@@ -12,18 +17,28 @@ import io
 
 from PIL import Image, UnidentifiedImageError
 
-from gamatrix.constants import PROFILE_PIC_SIZE
+from gamatrix.constants import PROFILE_PIC_MAX_INPUT_PIXELS, PROFILE_PIC_SIZE
+
+# Arm Pillow's built-in bomb detector at our own threshold: load() raises
+# DecompressionBombError past 2x this, and warns past it, regardless of format.
+Image.MAX_IMAGE_PIXELS = PROFILE_PIC_MAX_INPUT_PIXELS
 
 
 def process_profile_pic(raw: bytes) -> bytes:
     """Validate, resize, and re-encode an uploaded image as a PNG thumbnail.
 
-    Raises ValueError if the bytes aren't a readable image.
+    Raises ValueError if the bytes aren't a readable image or describe a bitmap
+    larger than PROFILE_PIC_MAX_INPUT_PIXELS.
     """
     try:
         opened = Image.open(io.BytesIO(raw))
+        # open() only reads the header, so size is known before any pixels are
+        # decoded. Reject oversized bitmaps here, before the expensive load().
+        width, height = opened.size
+        if width * height > PROFILE_PIC_MAX_INPUT_PIXELS:
+            raise ValueError("That image is too large to process.")
         opened.load()
-    except (UnidentifiedImageError, OSError) as exc:
+    except (UnidentifiedImageError, OSError, Image.DecompressionBombError) as exc:
         raise ValueError("That file isn't a readable image.") from exc
 
     img = opened.convert("RGBA")
--- a/src/gamatrix/images.py
+++ b/src/gamatrix/images.py
@@ -1,0 +1,35 @@
+"""Profile-picture image processing.
+
+User-uploaded images are normalized server-side before they reach S3: decoded
+(which validates that the bytes really are an image), flattened onto a
+transparent canvas, downscaled to a small square-ish thumbnail, and re-encoded
+as PNG. Re-encoding also strips any EXIF/metadata the original carried.
+"""
+
+from __future__ import annotations
+
+import io
+
+from PIL import Image, UnidentifiedImageError
+
+from gamatrix.constants import PROFILE_PIC_SIZE
+
+
+def process_profile_pic(raw: bytes) -> bytes:
+    """Validate, resize, and re-encode an uploaded image as a PNG thumbnail.
+
+    Raises ValueError if the bytes aren't a readable image.
+    """
+    try:
+        opened = Image.open(io.BytesIO(raw))
+        opened.load()
+    except (UnidentifiedImageError, OSError) as exc:
+        raise ValueError("That file isn't a readable image.") from exc
+
+    img = opened.convert("RGBA")
+    # thumbnail() preserves aspect ratio, fitting within the bounding box.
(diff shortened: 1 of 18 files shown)


src/gamatrix/preferences.py
```diff
@@ -91,3 +96,51 @@ async def save_profile(
...
+
+
+@router.post("/profile/pic")
+async def upload_profile_pic(
+    file: UploadFile = File(...),
+    user: dict = Depends(current_user_api),
+    repo: Repository = Depends(get_repo),
+):
+    """Accept an image upload, resize it, and store it as the user's pic."""
+    raw = await file.read()
```
derek-keeler: Had to pull out `gipity` on this one! I did notice that the entire file is read before any bounds checking is done. That means if I as `jerkface_usah` upload a massive file this routine will read the entire thing before it even looks at the size. Turns out my head is _also_ good as a hatrack, not _only_.

> oversized uploads are fully buffered before rejection
>
>line 108-113 reads the entire file, then checks `len(raw)`. An attacker could force the app to accept the whole multipart body into Starlette’s spool/temp storage before being rejected. This could provide some container-disk / memory pressure that might result in a failure state. Attempt to change to chunked reads with an early abort once the total crosses PROFILE_PIC_MAX_UPLOAD_SIZE.

src/gamatrix/images.py
```diff
@@ -0,0 +1,35 @@
...
+def process_profile_pic(raw: bytes) -> bytes:
+    """Validate, resize, and re-encode an uploaded image as a PNG thumbnail.
+
+    Raises ValueError if the bytes aren't a readable image.
+    """
+    try:
+        opened = Image.open(io.BytesIO(raw))
+        opened.load()
+    except (UnidentifiedImageError, OSError) as exc:
+        raise ValueError("That file isn't a readable image.") from exc
```
derek-keeler, line 42: Another find by `gipity`:

> image decompression-bomb / memory-exhaustion risk
>
>24-31 does Image.open(...); opened.load(); ... convert("RGBA") with no pixel-count guard. The 5 MB input cap does not protect you here: a highly-compressed PNG/WebP can be small on disk but enormous once decoded. That can spike CPU and RAM during load() or convert() and take down a worker. This path should enforce a hard max on dimensions / total pixels and reject Pillow decompression-bomb warnings/errors before fully decoding.

...if possible, you might want to do a 'sanitize_check' function that does as much checking as possible before calling `Image.open`.

src/gamatrix/preferences.py
```diff
@@ -91,3 +96,51 @@ async def save_profile(
...
+
+    key = f"profile_img/{user['email'].lower()}.png"
+    get_s3().put_bytes(key, png, "image/png")
+    updated = int(time.time())
+    repo.update_user(user["email"], {"pic_key": key, "pic_updated": updated})
+    url = pic_url({**user, "pic_key": key, "pic_updated": updated})
+    return JSONResponse({"pic_url": url})
+
+
+@router.get("/profile_img/{user_id}")
```
derek-keeler, line 144: One kind-issue: are profile images truly behind auth? (kinda?)

> authenticated image responses are marked public
>
> lines 127-145 gates /profile_img/{user_id} behind login, but then returns `Cache-Control: public, max-age=86400`. That invites shared proxy/CDN caching of an authenticated response. If this route is meant to stay auth-gated, the cache header should be private; otherwise the image should be moved to a truly public asset path/object URL.

...my 2¢ is to store images in public asset path, but I leave that decision to you... (not sure if the infra is set up for that).

derek-keeler (approved):
Some security-adjacent issues here, and one perf note. 

Security-adjacent comments are inline in the code changes.

**Perf note**: /profile_img/{user_id} calls repo.get_user_by_user_id(), which scans the whole users table each time (src/gamatrix/storage/dynamo.py:98-103). Since the branch now renders profile pics in multiple places, that adds avoidable per-image lookup cost on first render. Might be able to do something about this with local browser cache? I'm not an expert on how these things are handled...

eniklas:
Thanks @derek-keeler (and `gipity`) — all four addressed in the latest push:

1. **Buffered-before-size-check.** Confirmed the multipart path was the real problem: FastAPI/Starlette spools the *entire* body (to disk past 1MB) before the handler runs, so an in-handler chunked read would've been theater. Switched the upload to a **raw request body** read via `request.stream()` in capped chunks, aborting the moment it crosses 5MB — nothing buffers in full first. Frontend now POSTs the file as the raw body instead of multipart.

2. **Decompression bomb.** `process_profile_pic` now rejects on the header's declared dimensions *before* `load()`, and arms Pillow's own bomb detector (`Image.MAX_IMAGE_PIXELS`) as a backstop. A small-on-disk / huge-when-decoded image is refused before it can spike CPU/RAM. Added a test that trips the guard.

3. **`public` cache on an auth-gated response.** Changed to `Cache-Control: private`. Kept it auth-gated rather than moving to a public asset path (see below).

4. **Perf note (scan per image).** Took your storage hint and went with **DynamoDB**: pics now live in a dedicated `profile_pics` table keyed by `user_id`, so serving is a single `get_item` instead of the full-table `get_user_by_user_id` scan. 128px PNGs are tens of KB, well under the 400KB item limit. This also drops S3 from the feature entirely. (Trade-off vs. a public S3/CDN object URL: that would make true shared caching safe but un-gates the images; I kept them behind login + browser cache for now.)

`just check` is green (56 tests). End-to-end verified against moto: 404 → 400 non-image → 400 oversized (early abort) → 200 upload → served 128px `image/png` with `private` cache.

src/gamatrix/images.py
derek-keeler: Oh, that's kind of neat how that works! All in the same block to keep context with the image. Very cool.

src/gamatrix/preferences.py
derek-keeler: Beauty. Now I can post those lewd images of my junk and only my authenticated friends can see it!

derek-keeler (approved):
Looks good - ship'r

PR #152, opened 2026-06-08
Guard `just deploy` against a missing/stale deploy config

eniklas:
## Why

If `cdk-config.yaml` is absent, `load_deploy_config()` returns an empty config and the stack synths **without a custom domain**. A `just deploy` in that state silently tears down the Route53 alias records and the ACM cert, and CloudFormation still reports `UPDATE_COMPLETE` because a domainless stack is a "valid" outcome.

This took `gamatrix.erikniklas.com` / `games.erikniklas.com` offline after a deploy where the `gamatrix-configs` repo was checked out to an old v1 commit that predated `cdk-config.yaml`.

## What

A `_check-deploy-config` preflight that `deploy` now depends on:

- **Hard-fails** if `cdk-config.yaml` is missing at `GAMATRIX_CONFIG_DIR` (default `../gamatrix-configs`), mirroring `config.py`'s resolution. This blocks the exact deploy that caused the outage.
- **Warns** (but proceeds) if the config repo isn't on `master`, including the detached-HEAD case behind this incident.

## Tested

- config present + on `master` → passes silently.
- missing file → `ERROR ...`, exit 1 (deploy aborts).
- detached HEAD → `WARNING: gamatrix-configs is on a detached HEAD, not master.`, exit 0 (continues).

### DIFF
--- a/justfile
+++ b/justfile
@@ -56,8 +56,33 @@ build:
   docker build --build-arg SETUPTOOLS_SCM_PRETEND_VERSION={{version}} --target lambda-enricher -t {{container_name}}-enricher:{{version}} .
   docker build --build-arg SETUPTOOLS_SCM_PRETEND_VERSION={{version}} --target lambda-parser -t {{container_name}}-parser:{{version}} .
 
+# Fail fast if the private deploy config is missing, warn if it's off master.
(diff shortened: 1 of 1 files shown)


derek-keeler (approved):
Good idea.

eniklas:
> Good idea.

Experience is a tough teacher.

PR #153, opened 2026-06-09
Add Passkey login functionality

derek-keeler:
## What

Add passkey logins for gamatrix. Let's leave passwords in the dust.

- Added end-to-end passkey support so users can sign in with a passkey and manage their passkeys from within Gamatrix.
- Built the supporting backend, storage, and infrastructure needed to safely register, verify, store, and use passkeys in both local and deployed environments.
- Polished the experience with security hardening, clearer UX, stronger documentation, and broader automated test coverage.

## How

- Added passkey registration and sign-in flows to the authentication system, including browser-facing routes, WebAuthn ceremony handling, and account-level passkey management.
 - Introduced persistent storage and challenge tracking for passkeys, plus the required application and infrastructure configuration so the feature works with the app’s existing FastAPI, DynamoDB, and deployment setup.
 - Updated the UI so users can start passkey sign-in from the login page, manage passkeys from preferences, get clearer guidance during setup, and land on a simpler confirmation view after adding a passkey.
 - Strengthened the backend validation to better prevent replayed sign-in attempts, added the missing documentation expected by the repo conventions, and expanded tests to cover the new auth, management, and infrastructure behavior.

## Checklist

- [x] PR checks pass

### DIFF
--- a/.env-sample
+++ b/.env-sample
@@ -22,6 +22,12 @@ EMAIL_FROM=noreply@example.com
 # Base URL used to build password-reset links in emails.
 APP_BASE_URL=http://localhost:8088
 
+# WebAuthn relying-party scope. Production must use the stable canonical HTTPS
+# domain and explicitly list its exact origin.
+WEBAUTHN_RP_ID=localhost
+WEBAUTHN_RP_NAME=Gamatrix
+WEBAUTHN_ORIGINS=["http://localhost:8088"]
+
 # How long (days) before cached IGDB data is considered stale and re-enriched.
 IGDB_STALE_DAYS=30
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
     - name: Install dependencies
       # --frozen installs the exact versions in uv.lock and fails if the lock is
       # out of sync with pyproject.toml, so CI matches local and deployed builds.
-      run: uv sync --frozen --extra dev --extra ci
+      run: uv sync --frozen --extra dev --extra ci --extra cdk
 
     - name: Analyze with mypy
       run: uv run mypy
--- a/infrastructure/cdk/README.md
+++ b/infrastructure/cdk/README.md
@@ -31,14 +31,15 @@ npx cdk deploy
 ## What it creates
 
 - DynamoDB tables: `games`, `users`, `user_libraries` (+ `release_key-index` GSI),
-  `enrichment_jobs`, `metadata_overrides`, `config` (PAY_PER_REQUEST, PITR on)
+  `enrichment_jobs`, `metadata_overrides`, `config`, `passkeys`
+  (+ `user_handle-index` GSI), and TTL-enabled `auth_challenges`
+  (PAY_PER_REQUEST, PITR on)
 - S3 upload bucket (1-day lifecycle expiry, CORS for browser POST)
 - SQS enrichment queue
 - Lambdas: web (HTTP API + Mangum), enricher (SQS-triggered), parser (S3-triggered)
 - HTTP API (API Gateway v2) in front of the web Lambda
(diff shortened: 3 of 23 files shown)


eniklas (commented):
## Code review

Reviewed the diff, checked out the branch, and ran the suite locally. Verdict up front: **approve with minor follow-ups** — this is a well-structured, security-conscious WebAuthn implementation with strong test coverage.

**Local verification (branch `passkey`):**
- `pytest` — 92 passed
- `black --check` — clean
- `flake8 src tests` — clean
- `mypy` — no issues in 33 source files

### Strengths

- **Layered replay protection, both atomic.** `consume_auth_challenge` uses a conditional delete and `update_passkey_after_authentication` gates on `sign_count = :expected`; both are explicitly tested (`test_challenge_consumption_is_atomic`, `test_nonzero_counter_update_rejects_stale_concurrent_write`).
- **Idempotent user-handle assignment** via `if_not_exists(webauthn_user_id, :v)` — safe under concurrency.
- **Defense in depth** — registration and deletion re-verify the password; `exclude_credentials` blocks double-registration; resident key + UV required for discoverable passkeys.
- **Fail-loud config** — both the CDK stack and the `Settings` validator reject insecure production setups (non-HTTPS origins / `localhost` RP ID), with tests covering both guards.
- **No key material reaches templates** — asserted in `test_listing_hides_key_material_and_deletion_requires_password`.

### Issues & risks

**1. Canonical-domain middleware — redirect-loop risk behind API Gateway/Lambda (low, was my main concern; mostly mitigated).**
`app.py:_canonical_domain` compares `request.url.hostname` to `urlparse(settings.app_base_url).hostname` and 308-redirects on mismatch. I traced the production wiring: `gamatrix_stack.py` now requires a custom domain and sets *both* `APP_BASE_URL=https://{site_domain}` and `WEBAUTHN_RP_ID={site_domain}`. Under Mangum, `request.url.hostname` comes from the `Host` header, which for a custom domain on an HTTP API equals `site_domain` — so legitimate canonical requests won't loop, and raw `execute-api` hostnames correctly redirect to canonical (the intent). The residual risk is only if API Gateway ever surfaces an unexpected `Host` header. Note this branch is *untested* — the suite sets `LOCAL_DEV=true`, which skips the middleware entirely. Worth a single test that exercises the non-local branch with a mismatched host, and a smoke check on first deploy.

**2. `app_base_url` default vs `.env-sample` mismatch (low).**
`config.py` defaults `app_base_url` to `http://localhost:8080`, while `.env-sample`/`WEBAUTHN_ORIGINS` use `:8088`. This PR makes `app_base_url` load-bearing for redirect behavior, so aligning the two avoids a surprising local foot-gun.

**3. Zero-counter authenticators get no counter-based replay protection (inherent).**
Synced passkeys often report `sign_count == 0` forever, so the `sign_count = :expected` condition is always satisfied. This is an accepted WebAuthn limitation and the single-use challenge is the real anti-replay control here — a one-line comment saying so would help future readers.

### Minor / nits

- `routes.py`: `raise _passkey_error(exc)` drops the original `PasskeyError` cause from the HTTP path; `... from exc` would preserve it. Cosmetic.
- `delete_passkey` returns a bare 404 (no `detail`) for both "no user handle" and "conditional delete failed" — indistinguishable to clients.
- Client logic in `passkeys.js` (encode/decode, conditional mediation) has no behavioral test; the Python route tests only assert key strings are present. Acceptable given the size.

Nice work — the ceremony logic, atomicity, and config guards are sound.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

derek-keeler:
Awesome thanks! I'll update the PR tomorrow. (And by "I'll" I mean I'll make one of the bots do it).

eniklas (approved):
Faaaaancy 🚀

PR #154, opened 2026-06-09
Fix DB updated tooltip always showing 'never'

derek-keeler:
## What

Fix for the DB updated-on mouse-over

## How

The ingest path stamped db_updated_at on each user_libraries row but never on the user record itself, so the main-page tooltip always fell back to 'never'. Now writes the same timestamp to the users table via get_user_by_user_id + update_user after each successful ingest.

## Checklist

- [x] PR checks pass


eniklas (approved):
Nice one!

PR #155, opened 2026-06-09
Fix "Refresh missing IGDB" silently skipping all games

eniklas:
## What

The **Refresh missing IGDB** button currently updates **no titles**. This fixes it.

## Why

`refresh_missing` (`games/routes.py`) collects games with `enrichment_status == "not_found"` and enqueues an enrichment job for them — but it never changes their stored status.

Meanwhile the enricher skips any game whose status isn't `None` or `"pending"`:

```python
if game.get("enrichment_status") not in (None, ENRICHMENT_PENDING):
    continue
```

That guard was added in #134 ("Fix igdb retries") to stop redelivered/concurrent jobs from re-enriching already-`done` games — but it also means every `not_found` game queued by `refresh_missing` is filtered straight back out. `refresh_all` works because it explicitly marks each game `pending` before enqueuing; `refresh_missing` was never updated to match, so it regressed silently.

## How

- Flip the targeted `not_found` games to `pending` before enqueuing, reusing the records already returned by `scan_all_games()` (no extra reads), mirroring `refresh_all`.
- Add `tests/test_refresh_igdb.py` covering the status flip, that only `not_found` games are queued, and the no-op case. The status-flip test fails on `master` and passes with this change.

## Checklist

- [x] PR checks pass (local: `pytest` 82 passed, `black`, `flake8`, `mypy` all clean)
- [x] Tests added for the change
- [ ] Version label — patch (bug fix), no label needed

### DIFF
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -193,11 +193,16 @@ def refresh_missing(
     repo: Repository = Depends(get_repo),
 ):
     """Re-enrich games IGDB didn't find last time."""
-    release_keys = [
-        g["release_key"]
+    missing = [
+        g
         for g in repo.scan_all_games()
         if g.get("enrichment_status") == ENRICHMENT_NOT_FOUND
     ]
+    # The enricher only touches games that are unenriched or pending, so flip
+    # these from not_found to pending or they'd be skipped (see #134).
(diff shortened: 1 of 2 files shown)


derek-keeler (approved):
Easy schmeazy

PR #159, opened 2026-06-10
Filter Game Pass titles GOG flags as owned (#120)

eniklas:
## Problem

This filters out more Xbox titles that are no longer owned; the full fix needs to happen in the GOG integration.

The v2 #120 fix (`get_subscription_release_keys`) excludes Xbox titles only when `LicensedReleases.isOwned = 0`. But on my current DB, **42 Xbox titles survive that filter while only 1 is genuinely owned**.

The cause: `isOwned` is unreliable. All 42 survivors have `isOwned = 1`, including obvious Game Pass titles — Starfield, Sea of Thieves, Hades, Persona 5 Royal, Gears of War 4, etc. GOG never flips the flag when those leave the library/subscription.

## Fix

GOG also tracks subscription membership in `SubscriptionReleases` (each Game Pass tier → `LicensedReleases.libraryId`). 21 of the 42 survivors are listed there despite `isOwned = 1`. This change excludes a release if **either** signal says subscription:

```sql
WHERE lic.isOwned = 0 OR sr.licenseId IS NOT NULL
```

On my DB this drops the false-positive Xbox count from **42 → 21**.

## Known limitation

The remaining 21 are titles that left Game Pass *entirely* (Hades, Persona 5 Royal, The Outer Worlds, Dead Cells, …). GOG removes them from `SubscriptionReleases` but keeps `isOwned = 1` and their `ProductPurchaseDates` rows, leaving them byte-for-byte indistinguishable from genuine purchases — no DB field separates them from the one truly-owned title. Filtering those would require a manual per-user exclude list (not in this PR).

## Testing

- Added a test case: an Xbox title with `isOwned = 1` that is present in `SubscriptionReleases` is now excluded.
- `83 passed` locally; verified against a real post-subscription DB (772 excluded, 21 kept).

### DIFF
--- a/src/gamatrix/gogdb/parser.py
+++ b/src/gamatrix/gogdb/parser.py
@@ -61,21 +61,34 @@ class GogDBParser:
         ).fetchone()[0]
 
     def get_subscription_release_keys(self) -> set[str]:
-        """Release keys present only via a subscription (issue #120).
+        """Release keys available via a subscription, not owned (issue #120).
 
         Xbox Game Pass titles stay in ProductPurchaseDates after they leave the
-        subscription, so v1 reported them as owned forever. LicensedReleases
-        flags genuine ownership: isOwned = 0 means the title is only available
-        through a subscription (verified: all such rows are Xbox Game Pass).
+        subscription, so v1 reported them as owned forever. Two GOG tables flag
+        them, and neither is sufficient on its own:
+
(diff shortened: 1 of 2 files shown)


derek-keeler (approved):
I believe it looks just fine. The test makes it all make sense and the change is nice and simple.

PR #160, opened 2026-06-11
Count unique games in the header, not table rows

eniklas:
## Problem

The results header showed the number of table **rows**, not the number of unique games. In the grid view the same title can occupy more than one row when platform copies have different owners — those rows don't merge (the merge key is `(slug, owners)`), so a single game was counted twice.

## Fix

Compute the header total from **distinct slugs** among the displayed games instead of `len(games)`. Rows are already grouped by slug, so distinct slugs is the unique-game count.

```python
total = len({g["slug"] for g in games})
```

This flows through `build_caption()` into the `<caption>` header (both normal and grid views) and the "Random game selected from N" text. Row rendering is unchanged — two rows still display when ownership differs; only the count reflects unique games.

## Testing

- Added `test_count_is_unique_games_not_rows`: one title owned by different users on different platforms produces 2 rows but `result.total == 1`.
- Full suite: 104 passed; black / flake8 / mypy clean.

### DIFF
--- a/src/gamatrix/games/service.py
+++ b/src/gamatrix/games/service.py
@@ -89,7 +89,11 @@ def compare(repo: Repository, opts: CompareOptions) -> CompareResult:
         reverse=opts.direction == "desc",
     )
 
-    total = len(games)
+    # Count unique games, not rows: the grid view can list the same title on
+    # more than one row when platform copies have different owners, but those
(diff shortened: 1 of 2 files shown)


derek-keeler (approved):

derek-keeler:
I like it. Ship.

PR #161, opened 2026-06-12
Show DB upload time in the viewer's local time zone

eniklas:
## What

Hovering over a user in the filter list shows their DB upload time. It was rendered as the raw stored UTC ISO string (e.g. `DB updated: 2026-06-11T14:32:00+00:00`). This converts it to the viewer's local time zone, e.g. `6/11/2026, 7:32:00 AM`.

## How

The server can't know the viewer's time zone, but the browser can, so the conversion happens client-side:

- `db_updated_at` (an unambiguous UTC instant) is rendered into a `data-db-updated` attribute on the user `<label>`.
- A small inline script rewrites each tooltip with `new Date(iso).toLocaleString()`, which automatically uses the browser's locale and time zone — no TZ plumbing or stored preference needed.

The original server-rendered `title` stays in place as a no-JS fallback. The filter form renders only once (HTMX swaps only `#games-container`), so a single pass on page load covers every label.

## Testing

- `uv run pytest tests/test_gogdb.py tests/test_games.py` — passes (15).
- Jinja template parse-checked.

No backend or data changes.

### DIFF
--- a/src/gamatrix/templates/games.html.jinja
+++ b/src/gamatrix/templates/games.html.jinja
@@ -27,7 +27,10 @@
     <fieldset class="filter-group">
         <legend>Users</legend>
         {% for uid, u in users.items() %}
-        <label title="DB updated: {{ u.get('db_updated_at', 'never') }}">
+        {# Server stores db_updated_at as UTC; the title here is a no-JS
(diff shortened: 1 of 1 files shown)


derek-keeler (approved):
Ha! was just going to fix this... nice.

PR #162, opened 2026-06-13
Add personal API tokens for unattended DB uploads (#129)

### LINKED ISSUE #129: Add unattended DB uploads
v1 allowed DB uploads with `curl`, which could be done as a scheduled task. This functionality was lost in v2.0.0 because authentication is required. Come up with a way to do this in v2.

eniklas:
Closes #129.

## Problem

v1 let a scheduled task `curl` the GOG Galaxy DB to an unauthenticated endpoint. v2 lost this: every upload path is gated by a session cookie that only an interactive browser login (password or passkey) produces. There was no non-interactive credential a scheduled task could use.

In AWS the scripted flow actually only needs **one** authenticated call — `GET /upload/presign`. Once the script has the presigned POST it uploads straight to S3, and the existing `OBJECT_CREATED` event triggers `db_parser`, which derives the account from the `uploads/<email>.db` key and ingests. (`/upload/complete` is already a no-op in AWS.) So the whole problem reduces to "how does a headless script authenticate to `/upload/presign`."

## Approach — personal API tokens

Revocable, upload-scoped bearer tokens, one per machine.

- **`api_tokens` table** (PK `token_id`, GSI `email-index`). Token format `gmx_<token_id>_<secret>`; only `sha256(secret)` is stored, compared in constant time, so a DB leak can't be replayed. Lookup is O(1) by `token_id`; the GSI lists a user's tokens for the management UI.
- **`current_user_upload` dependency** accepts `Authorization: Bearer <token>` in addition to the session cookie, and is attached **only** to the upload routes — a token can't substitute for a cookie anywhere else in the app.
- **Token UI** at `/auth/tokens` (linked from preferences and the upload page): create (re-authenticated with the account password, like passkey management), list with created/last-used, and revoke. On creation the secret is shown **once** alongside a ready-to-paste PowerShell snippet, prefilled with the token and site URL, that stores the token with locked-down NTFS perms (`icacls`), downloads the uploader, and registers a daily scheduled task.
- **Reference uploaders** served from `/static`:
  - `upload-gamatrix.ps1` — Windows. Uses the in-box `curl.exe` for the S3 multipart upload and `Invoke-RestMethod` for the presign, so it runs on a **stock Windows 10/11 with Windows PowerShell 5.1** — no Python and no PowerShell 7 needed. Copies the locked `galaxy-2.0.db` with shared-read access first.
  - `upload-gamatrix.sh` — macOS/Linux (bash + curl).

## Alternatives considered

- **Long-lived JWT "personal access token"** (`POST /auth/token` returning a long-TTL JWT): reuses existing JWT machinery, but JWTs can't be revoked without a denylist and still expire — weaker than a stored, revocable token for little less effort.
- **Script replays the browser login** (POST `/auth/login`, capture the cookie): zero backend change, but it stores the **account password** on disk and breaks if the account goes passkey-only or rotates its password.
- **Scoped S3 credentials / per-user IAM** (`s3:PutObject` on `uploads/<email>.db`, bypass the app): no app auth at all, but distributing and rotating AWS creds per user doesn't scale and is awkward for a friends-and-family app.
- **Client language: pure `.bat`** was rejected — the presign step returns JSON, which batch can't parse cleanly. **Pure PowerShell without `curl.exe`** was rejected because `Invoke-RestMethod -Form` only exists in PowerShell 7; leaning on the in-box `curl.exe` keeps the client dependency-free on a default Windows install.

## Token storage & permissions guidance (shown in the UI)

The reveal panel and the generated snippet tell the user to keep the token out of the script and locked down:
- **Windows:** saved to `%USERPROFILE%\.gamatrix\token`, then `icacls /inheritance:r /grant:r "<user>:(R)"` so only that account can read it.
- **macOS/Linux:** `~/.gamatrix-token` with `chmod 600` (or the `GAMATRIX_TOKEN` env var).

The uploaders never take the token on the command line by default (it would leak into the process list / task definition); they read it from that file or env var.

## Tests

`tests/test_tokens.py` covers the token service (roundtrip, last-used stamping, garbage/tampered/wrong-secret rejection), owner-scoped revocation, the create/list/revoke routes (incl. password re-auth and 404s), and the core of #129: `/upload/presign` authenticating with a bearer token and no cookie, rejecting a bad token, and still working with a cookie.

`just check` is green: 114 passed, black clean, mypy clean. CDK, `scripts/init_local.py`, and `tests/conftest.py` all create the new table + GSI.

### DIFF
--- a/src/gamatrix/auth/routes.py
+++ b/src/gamatrix/auth/routes.py
@@ -6,10 +6,11 @@ from fastapi import APIRouter, Depends, Form, HTTPException, Request, status
 from fastapi.responses import HTMLResponse, JSONResponse, RedirectResponse
 from pydantic import BaseModel
 
-from gamatrix.auth import passkeys, service
+from gamatrix.auth import passkeys, service, tokens
 from gamatrix.auth.dependencies import current_user, current_user_api, get_repo
 from gamatrix.auth.service import COOKIE_NAME
 from gamatrix.config import get_settings
+from gamatrix.constants import API_TOKEN_NAME_MAX_LENGTH
 from gamatrix.storage.dynamo import Repository
 from gamatrix.templating import templates
 
@@ -27,6 +28,15 @@ class CeremonyVerificationRequest(BaseModel):
 
 
 class DeletePasskeyRequest(BaseModel):
+    password: str
+
+
+class CreateTokenRequest(BaseModel):
+    password: str
+    name: str
+
+
+class DeleteTokenRequest(BaseModel):
     password: str
 
 
@@ -180,6 +190,95 @@ def delete_passkey(
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
             detail="Passkey not found for this account.",
+        )
+    return {"deleted": True}
+
+
+def _token_setup_snippet(token: str) -> str:
+    """A ready-to-paste PowerShell block: store the token with locked-down
+    permissions, fetch the uploader, and schedule a daily run."""
+    base_url = get_settings().app_base_url.rstrip("/")
+    return (
+        "# 1) Save your token where only your Windows account can read it.\n"
+        '$dir = "$env:USERPROFILE\\.gamatrix"\n'
+        "New-Item -ItemType Directory -Force $dir | Out-Null\n"
+        f"Set-Content \"$dir\\token\" '{token}' -NoNewline\n"
+        "# Strip inherited permissions, then grant read to just you:\n"
+        'icacls "$dir\\token" /inheritance:r /grant:r "$($env:USERNAME):(R)" '
+        "| Out-Null\n\n"
+        "# 2) Download the uploader script.\n"
+        f'Invoke-WebRequest "{base_url}/static/upload-gamatrix.ps1" '
+        '-OutFile "$dir\\upload-gamatrix.ps1"\n\n'
+        "# 3) Schedule a daily upload at 5am (adjust the time as you like).\n"
+        "$action = New-ScheduledTaskAction -Execute powershell.exe -Argument "
+        f'"-ExecutionPolicy Bypass -File `"$dir\\upload-gamatrix.ps1`" '
+        f'-BaseUrl {base_url}"\n'
+        "$trigger = New-ScheduledTaskTrigger -Daily -At 5am\n"
+        "Register-ScheduledTask -TaskName 'Gamatrix DB upload' -Action $action "
+        "-Trigger $trigger\n"
+    )
+
+
+@router.get("/tokens", response_class=HTMLResponse)
+def token_management(
+    request: Request,
+    user: dict = Depends(current_user),
+    repo: Repository = Depends(get_repo),
+):
+    return templates.TemplateResponse(
+        request,
+        "tokens.html.jinja",
+        {
+            "user": user,
+            "tokens": repo.list_api_tokens(user["email"]),
+            "base_url": get_settings().app_base_url.rstrip("/"),
+        },
+    )
+
+
+@router.get("/tokens/list", response_class=HTMLResponse)
+def list_tokens(
+    request: Request,
+    user: dict = Depends(current_user),
+    repo: Repository = Depends(get_repo),
+):
+    return templates.TemplateResponse(
+        request,
+        "tokens_list.html.jinja",
+        {"tokens": repo.list_api_tokens(user["email"])},
+    )
+
+
+@router.post("/tokens")
+def create_token(
+    body: CreateTokenRequest,
+    user: dict = Depends(current_user_api),
+    repo: Repository = Depends(get_repo),
+):
+    if not service.authenticate(repo, user["email"], body.password):
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN, detail="Wrong password."
+        )
+    name = body.name.strip()[:API_TOKEN_NAME_MAX_LENGTH] or "Unnamed token"
+    token = tokens.create_api_token(repo, user["email"], name)
+    return {"token": token, "snippet": _token_setup_snippet(token)}
+
+
+@router.delete("/tokens/{token_id}")
+def delete_token(
+    token_id: str,
+    body: DeleteTokenRequest,
+    user: dict = Depends(current_user_api),
+    repo: Repository = Depends(get_repo),
+):
+    if not service.authenticate(repo, user["email"], body.password):
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN, detail="Wrong password."
+        )
+    if not repo.delete_api_token(token_id, user["email"]):
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail="Token not found for this account.",
         )
     return {"deleted": True}
 
--- a/src/gamatrix/auth/routes.py
+++ b/src/gamatrix/auth/routes.py
@@ -201,11 +201,22 @@ def _token_setup_snippet(token: str) -> str:
     return (
         "# 1) Save your token where only your Windows account can read it.\n"
         '$dir = "$env:USERPROFILE\\.gamatrix"\n'
+        '$tokenPath = "$dir\\token"\n'
         "New-Item -ItemType Directory -Force $dir | Out-Null\n"
-        f"Set-Content \"$dir\\token\" '{token}' -NoNewline\n"
-        "# Strip inherited permissions, then grant read to just you:\n"
-        'icacls "$dir\\token" /inheritance:r /grant:r "$($env:USERNAME):(R)" '
-        "| Out-Null\n\n"
+        f"Set-Content $tokenPath '{token}' -NoNewline\n"
+        "# Replace the file's ACL outright (no inherited perms) and grant just\n"
+        "# you read+write, so you can re-cycle the token later without a fight.\n"
+        "$currentUser = "
+        "[System.Security.Principal.WindowsIdentity]::GetCurrent().Name\n"
+        "$acl = New-Object System.Security.AccessControl.FileSecurity\n"
+        "$acl.SetAccessRuleProtection($true, $false)\n"
+        "$rule = New-Object System.Security.AccessControl.FileSystemAccessRule("
+        "$currentUser, "
+        "([System.Security.AccessControl.FileSystemRights]::Read -bor "
+        "[System.Security.AccessControl.FileSystemRights]::Write), "
+        "[System.Security.AccessControl.AccessControlType]::Allow)\n"
+        "$acl.SetAccessRule($rule)\n"
+        "Set-Acl -Path $tokenPath -AclObject $acl\n\n"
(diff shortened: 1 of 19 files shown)


src/gamatrix/auth/routes.py
```diff
@@ -184,6 +194,95 @@ def delete_passkey(
...
+        f'Invoke-WebRequest "{base_url}/static/upload-gamatrix.ps1" '
+        '-OutFile "$dir\\upload-gamatrix.ps1"\n\n'
+        "# 3) Schedule a daily upload at 5am (adjust the time as you like).\n"
+        "$action = New-ScheduledTaskAction -Execute powershell.exe -Argument "
+        f'"-ExecutionPolicy Bypass -File `"$dir\\upload-gamatrix.ps1`" '
+        f'-BaseUrl {base_url}"\n'
+        "$trigger = New-ScheduledTaskTrigger -Daily -At 5am\n"
+        "Register-ScheduledTask -TaskName 'Gamatrix DB upload' -Action $action "
+        "-Trigger $trigger\n"
+    )
```
derek-keeler, line 230: I think the intent here is to do a `chmod a-rwx,u+r` on the `~/.gamatrix/token` file alone, right? I'd probably suggest doing `chmod a-rwx,u+rw` to allow the user to easily update their token anytime they cycle it (for whatever reason).

Windows ACEs (ACLs? What happened to that?) are scary and I don't understand the model fully so I asked copilot to walk me through the plan here to see what nuance there might have been. I also gave it the parameter of "explain to me like I'm a weekend Linux admin wannabe" so I'll paraphrase it's results as best I can for you here.

Suggested alteration of the Windows secure-token PowerShell snippet:


```suggestion
 return (
     "# 1) Save your token where only your Windows account can read it.\n"
     '$dir = "$env:USERPROFILE\\.gamatrix"\n'
     '$tokenPath = "$dir\\token"\n'
     "New-Item -ItemType Directory -Force $dir | Out-Null\n"
     f"Set-Content $tokenPath '{token}' -NoNewline\n"
     "# Fully replace the file ACL with the Windows equivalent of a per-user token file.\n"
     "$currentUser = [System.Security.Principal.WindowsIdentity]::GetCurrent().Name\n"
     "$acl = New-Object System.Security.AccessControl.FileSecurity\n"
     "$acl.SetAccessRuleProtection($true, $false)\n"
     "$rule = New-Object System.Security.AccessControl.FileSystemAccessRule("
     "$currentUser, "
     "([System.Security.AccessControl.FileSystemRights]::Read -bor "
     "[System.Security.AccessControl.FileSystemRights]::Write), "
     "[System.Security.AccessControl.AccessControlType]::Allow)\n"
     "$acl.SetAccessRule($rule)\n"
     "Set-Acl -Path $tokenPath -AclObject $acl\n\n"
     "# 2) Download the uploader script.\n"
     f'Invoke-WebRequest "{base_url}/static/upload-gamatrix.ps1" '
     '-OutFile "$dir\\upload-gamatrix.ps1"\n\n'
     "# 3) Schedule a daily upload at 5am (adjust the time as you like).\n"
     "$action = New-ScheduledTaskAction -Execute powershell.exe -Argument "
     f'"-ExecutionPolicy Bypass -File `"$dir\\upload-gamatrix.ps1`" '
     f'-BaseUrl {base_url}"\n'
     "$trigger = New-ScheduledTaskTrigger -Daily -At 5am\n"
     "Register-ScheduledTask -TaskName 'Gamatrix DB upload' -Action $action "
     "-Trigger $trigger\n"
 )
```

So the suggested change is basically this:
 
1. Use `chmod u+rw` instead of `chmod u+r` - Figuring out why I cannot delete `~/.gamatrix` will annoy Windows users later on. (And they'll have to figure out how to reset perms on that single file which is something I do once a decade or so...).
1. Fully _replace_ all permissions on the `token` file instead of _adding_ to it. `icacls` is an editor of the Windows ACEs, but doesn't really do wholesale replacements (kind of like using the `a-rwx` part of `chmod a-rwx,u+rw` when you want to fully change the perms on a file from scratch). Copilot's explainer: _This is closer to “build a fresh per-file ACL from scratch” than “mutate whatever permissions happened to be inherited.” In Linux terms, it’s more like explicitly setting the file’s final mode/ACL for the owner than doing a partial chmod on top of unknown existing state._

I checked to ensure this works as I expect on my local machine, and it totally clobbers the details on the `Security` pane (right-click on the file in Explorer and click 'Properties' then choose the 'Security' tab - do it once before the `Set-Acl` call and then once after.)

I also checked that this would use the stock-installed version of PowerShell that ships as part of Windows 11. Should be good.

src/gamatrix/templates/tokens.html.jinja
```diff
@@ -0,0 +1,128 @@
...
+    document.getElementById("copy-token").addEventListener("click", (e) =>
+        copy(tokenEl.textContent, e.currentTarget));
+    document.getElementById("copy-snippet").addEventListener("click", (e) =>
+        copy(snippetEl.textContent, e.currentTarget));
+
+    // Tokens load asynchronously, so delegate the revoke handler.
+    document.getElementById("token-list").addEventListener("click", async (event) => {
+        const button = event.target.closest(".delete-token");
+        if (!button) return;
+        const password = window.prompt("Confirm your current password to revoke this token:");
```
derek-keeler: This is a clear-text password popup no? That's probably something that should at least go into a `password` entry field. Would be best to make it inline how we have it elsewhere too - but that part is up to you.

src/gamatrix/static/upload-gamatrix.sh
```diff
@@ -0,0 +1,57 @@
...
+trap 'rm -f "$tmp"' EXIT
+# Galaxy holds the DB open; copying it is fine for a read.
+cp "$DB_PATH" "$tmp"
+
+# 1) Presign (the only authenticated call).
+presign="$(curl --fail --silent --show-error \
+  -H "Authorization: Bearer ${TOKEN}" \
+  "${BASE_URL%/}/upload/presign")"
+
+url="$(printf '%s' "$presign" | python3 -c 'import sys,json; print(json.load(sys.stdin)["url"])')"
```
derek-keeler, line 46: Might want to call out to users that a dependency is `python3` on the PATH. Not true on stock machines (Mac) and not _always_ true on Linux either. Also, folks that are "Windows with a side of Linux" would want to know this as well - not all WSL installations have `python3` available by default either.

I made a note of this in the HTML template as well...

src/gamatrix/static/upload-gamatrix.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# 2) Build -F args from the presign fields (file LAST; S3 ignores anything after it).
+mapfile -t fields < <(printf '%s' "$presign" | python3 -c '
```
derek-keeler: GPT-5.4 tells me that `mapfile` isn't a feature on bash that ships on stock MacOS, either.

GPT: `mapfile` _is not available in the old Bash 3.2 shipped by default on many macOS systems_
DDG search: _macOS 26 (Tahoe) ships with Bash version 3.2.57, which is considered deprecated._

So we might want to call that out too...

src/gamatrix/templates/tokens.html.jinja
```diff
@@ -0,0 +1,128 @@
...
+    <h4>Set it up on Windows (PowerShell)</h4>
+    <p class="muted">Paste this into a PowerShell window. It saves the token so
+    only your Windows account can read it, downloads the uploader, and schedules
+    a daily upload.</p>
+    <pre id="token-snippet" class="token-box"></pre>
+    <button type="button" id="copy-snippet">Copy setup script</button>
+    <p class="muted">On macOS or Linux, download
+    <a href="{{ base_url }}/static/upload-gamatrix.sh">upload-gamatrix.sh</a>,
+    save the token to a file only you can read
+    (<code>chmod 600 ~/.gamatrix-token</code>), and run the script from cron.</p>
```
derek-keeler: Might want to call out that the script relies on `python3` being on the PATH, and that for MacOS users `bash` must be updated to a recent supported version (Apple won't update from 3.2).

Made individual comments in your script calling out the dependencies...

derek-keeler (changes requested):
Looks good but there is a concern on Windows (ACEs) and some documentation tweaks. One must-fix (unless I didn't grok the usage right) is the password popup in the token generation flow.

derek-keeler:
I love that you did this - was thinking that was the next step for me too - my 9pm popup is still uploading to the OG gamatrix as we speak. As soon as this lands I'll update it.

src/gamatrix/auth/routes.py
eniklas: Applied your suggestion in acc3a7e — the snippet now builds a fresh `FileSecurity` with `SetAccessRuleProtection($true, $false)` and `Set-Acl`s it, replacing the inherited ACL wholesale rather than editing it with `icacls`. Grant is Read **and** Write (your `u+rw` point) so the token can be re-cycled and the `.gamatrix` dir stays deletable. Thanks for testing the Security pane behaviour on stock Win11 PowerShell.

src/gamatrix/templates/tokens.html.jinja
eniklas: Fixed in acc3a7e. The revoke flow no longer uses a clear-text `window.prompt` — clicking Revoke now reveals an inline masked `<input type="password">` (with `autocomplete="current-password"`) in the token row, matching the create-token form. (Worth noting the passkey-delete flow still uses `window.prompt` for the same re-auth; happy to give it the same treatment in a follow-up if you want consistency there.)

src/gamatrix/static/upload-gamatrix.sh
eniklas: Good call. The script header now spells out that `curl` **and** `python3` must be on the `PATH`, explicitly noting `python3` is not present on stock macOS, some Linux distros, or many WSL installs (acc3a7e).

src/gamatrix/static/upload-gamatrix.sh
eniklas: Fixed — dropped `mapfile` entirely in favour of a `while IFS= read -r` loop, so the script runs on Apple's bash 3.2 (acc3a7e). Added a comment in the header calling out that it's written for bash 3.2 for exactly this reason.

src/gamatrix/templates/tokens.html.jinja
eniklas: Done in acc3a7e — the tokens page now tells macOS/Linux users the script needs `curl` and `python3` on the `PATH` and to install `python3` if `python3 --version` fails. On the bash side I removed the `mapfile` use so it works on the stock 3.2 Apple ships, so no bash upgrade is required.

src/gamatrix/templates/tokens.html.jinja
derek-keeler: Oh wow! We missed that when I put the passkeys in... that's awful. Let's leave that one to me (I put that in there) and let you get this PR in place. Thanks for catching that!

derek-keeler (approved):
Thanks for fixing this up - ship'r

Issue #164, opened 2026-06-14
Improve local development workflows/processes


derek-keeler:
When iterating on UX and other issues it is very useful to have a working local version of the application that can be put up/taken down/refreshed quickly and easily.

It is highly useful to be able to generate "dummy" users and be able to use them to test various aspects of the application - but we do not want to check in binaries (especially large binaries) for test user-GOG-dbs.

Enhance our local dev workflow with a containerized solution that will allow for multiple test users being generated at will.

derek-keeler:
Completed and works on my machine. We should review together to make sure this works well for others too!

Issue #92, opened 2021-11-12
Investigate UX enhancements


derek-keeler:
Take a look at the various tutorials on Flask + React and see if we can build out a more expressive front end. This may be useful to help resolve issues with the front end such as #91 and #17 (and others) as we may be able to component-ize the front end with React elements.

eniklas:
@derek-keeler I'd say this is obsolete in v2, are we good to close this out?

derek-keeler:
Nope - this is coming. I have a prerequisite PR landing today (local dev enhancements) and then I'm ready to blast through this one.

The plan: Make local dev trivial (for some reason on my workstation the current flow leaves a bunch of zombie processes and updates things globally). Remove the need for 'user-dbs' repo, and make it trivial to add N test users.

Then: Abstract away the data layer and UX layer.

Then: Swap UX layer into a fully swappable template based scheme with 'modes' (light/dark/colour-blind).

It's a big'un.

PR #165, opened 2026-06-14
Enhance local development workflows with containerization and generated test data/test users

derek-keeler:
## What

Fix for #164 

Implementation of improvement of local workflows for developers working on gamatrix.

- Fully containerized local deployment.
- Generate test users using dev's local GOG db.
- Full application is stood up locally allowing for testing of all UX workflows.

## How

Addition of tooling and containerization logic to support the local dev workflow.

## Checklist

- [ ] PR checks pass

Note: This is a prerequisite to support meaningful work on #92.
Note2: This will also unlock UX-testing scenarios in Copilot/Claude Code, should that be interesting.


derek-keeler:
Note: during this work I discovered that Python 3.13 is supported, but 3.14 breaks.

eniklas (commented):
Overall this is a solid, well-tested PR. It replaces checked-in binary fixtures with a generator that derives git-ignored sample data from the developer's own GOG DB, adds a clean reset path, and documents the flow well. The fixture-generation logic (`compute_layout`/`assign_keys`/`build_fixture`) is sound — the ownership matrix math is correct and validated, and the schema-driven copy nicely tolerates upstream column additions. Test coverage for the new scripts and the dynamo purge helpers is good.

A few things to fix before merge:

### 🔴 `scripts/test_path.py` is debug cruft (should not be committed)
```python
"""Print the Python import path for local environment debugging."""
import sys
print(sys.path)
```
This is a stray debugging leftover with no caller and no test. Remove it.

### 🟠 dynamodb-local volume change breaks existing dev environments silently
`docker-compose.yml` now runs the container as the unprivileged `dynamodblocal` user (uid 1000) and mounts the `dynamodb-data` named volume at `/home/dynamodblocal`. The inline comment reasons correctly about a **fresh** volume inheriting the image dir's ownership — but a developer who already ran the old stack has a **pre-existing, root-owned** `dynamodb-data` volume. That volume keeps its root ownership, so the unprivileged process will fail to write the DB. These devs will need `docker volume rm gamatrix_dynamodb-data` (or `docker compose down -v`). Worth a one-line note in the README troubleshooting section, since the failure mode (permission denied on startup) is non-obvious.

### 🟡 Doc/comment contradictions
- **`justfile` (`seed-local` recipe comment)**: says *"(committed fixtures)"* — directly contradicting the PR's whole premise that fixtures are git-ignored and generated. Reword to "generated fixtures".
- **`docker-compose.yml`** comment says the service *"previously had to run as `user: root`"*, but `master` never had `user: root` (it's not in the diff and not in the current file). The referenced state only existed mid-branch, so the comment will confuse future readers. Consider trimming that parenthetical.

### Minor / non-blocking
- `bootstrap` runs `init-local` (which seeds config users from `seed_users.json`) immediately followed by `seed-local` (which hard-resets *all* users and seeds the manifest). The config-user seeding in `init-local` is therefore thrown away on the bootstrap path. Harmless and intentional given the recipe hardcodes `--hard-reset-existing-users`, but slightly wasteful.
- Default values (`games=20`, `common=5`, `pair=5`) are duplicated in both `generate_fixtures.py` and the `gen-fixtures` justfile recipe; they can drift. Letting the script own the defaults (omit the flags when unset, as already done for `users`/`usernames`) would be cleaner.

### Verified as correct
- `delete_all_passkeys` deletes solely on `credential_id`, which matches the passkeys table's key schema. ✓
- `create_user` accepts the `user_id`/`is_admin` kwargs `seed_sample_data.py` passes. ✓
- `dbPath="."` with WORKDIR `/home/dynamodblocal` resolves to the same volume root as the old `/data`, so the shared-db file location within the volume is unchanged (no data-location regression — only the ownership caveat above). ✓
- `moto` is already a dev dependency, so the new `mock_aws` test imports are fine. ✓

The two doc contradictions and `test_path.py` are quick fixes; the volume-ownership note is the one I'd most want addressed since it'll bite existing contributors on first pull.

eniklas (approved):

derek-keeler:
Updates complete - all raised issues have been resolved.

> ### 🔴 `scripts/test_path.py` is debug cruft (should not be committed)

- Removed - this was an intermediate test file left behind by Copilot that I manually checked in. (Don't waste tokens on checkin!)

> ### 🟠 dynamodb-local volume change breaks existing dev environments silently

- README updated with a warning in local-dev setup section about setup post v2.3.0.

> ### 🟡 Doc/comment contradictions

- Documentation coalesced and inconsistencies / drift corrected.

> ### Minor / non-blocking
>     * `bootstrap` runs `init-local` (which seeds config users from `seed_users.json`) immediately followed by `seed-local` (which hard-resets _all_ users and seeds the manifest). The config-user seeding in `init-local` is therefore thrown away on the bootstrap path. Harmless and intentional given the recipe hardcodes `--hard-reset-existing-users`, but slightly wasteful.

- Agreed - this was kind of clunky (and it still is a bit, admittedly). I kept the original 'seed_users' functionality for any dev that doesn't have a large enough GOGdb to mock up users with their own data. Doesn't apply to this group, but may be a problem for others. The documented/default workflow is to use the GOGdb-spawned test users.

>     * Default values (`games=20`, `common=5`, `pair=5`) are duplicated in both `generate_fixtures.py` and the `gen-fixtures` justfile recipe; they can drift. Letting the script own the defaults (omit the flags when unset, as already done for `users`/`usernames`) would be cleaner.

- Also agreed there is zero reason for the defaults to be repeated within the `justfile`. It can override defaults just fine if they are sent in.

derek-keeler:
@eniklas - I see you re-reviewed! Sorry - just pushed my last change (updating the default seed_users workflow) which is pretty trivial. Once that is accepted we can push these changes.

eniklas:
@derek-keeler you've got failing tests hoser.

eniklas (approved):
Nice one, thanks for making local dev sweet 😍

Issue #163, opened 2026-06-14
Passkey revocation flow requires a plain-text password popup


derek-keeler:
A plain-text popup appears when a user attempts to delete/revoke their passkey in the 'Preferences' screen. This popup isn't a proper password field or anything so the user types their password out in plain view on the screen.

Replace this with an inline flow (much like the user-token revocation flow from #162) so that we make the product more security-focused than it is and that we exercise good security hygiene throughout.

PR #167, opened 2026-06-14
Fix bug requesting users password from plain-text dialog in passkey revocation flow

derek-keeler:
Fix for #163 

Improvements to inline passkey revocation.

  - Replaced window.prompt with an inline HTMX confirmation form using a masked password field.
  - Added cancel, inline wrong-password feedback, and list refresh after deletion.
  - Enforced passkey ownership before exposing confirmation details.
  - Preserved the existing JSON delete endpoint.

## Checklist

- [ ] PR checks pass

NOTE: I have *not* tested this locally yet! (Waiting for #165 to land)...

### DIFF
--- a/src/gamatrix/auth/routes.py
+++ b/src/gamatrix/auth/routes.py
@@ -92,6 +92,22 @@ def _passkey_credentials(user: dict, repo: Repository) -> list[dict]:
     return repo.list_passkeys(user_handle) if user_handle else []
 
 
+def _owned_passkey(credential_id: str, user: dict, repo: Repository) -> dict:
+    user_handle = user.get("webauthn_user_id")
+    if not user_handle:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail="No passkeys are registered for this account.",
+        )
+    passkey = repo.get_passkey(credential_id)
+    if not passkey or passkey.get("user_handle") != user_handle:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail="Passkey not found for this account.",
+        )
+    return passkey
+
+
 @router.get("/passkeys", response_class=HTMLResponse)
 def passkey_management(
     request: Request,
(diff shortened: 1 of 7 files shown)


src/gamatrix/auth/routes.py
```diff
@@ -92,6 +92,22 @@ def _passkey_credentials(user: dict, repo: Repository) -> list[dict]:
     return repo.list_passkeys(user_handle) if user_handle else []
 
 
+def _owned_passkey(credential_id: str, user: dict, repo: Repository) -> dict:
```
eniklas, line 95: Good helper. Note the pre-existing `DELETE /auth/passkeys/{credential_id}` handler (line 243) still open-codes the same `webauthn_user_id` / ownership check instead of using this. Since this PR removes that endpoint's only caller (the `.delete-passkey` JS), consider either deleting that handler or refactoring it to call `_owned_passkey` so the two delete paths can't diverge.

src/gamatrix/auth/routes.py
```diff
@@ -118,6 +134,61 @@ def list_passkeys(
...
+):
+    return templates.TemplateResponse(
+        request,
+        "passkey_delete_form.html.jinja",
+        {"passkey": _owned_passkey(credential_id, user, repo), "error": None},
+    )
+
+
+@router.post("/passkeys/{credential_id}/delete", response_class=HTMLResponse)
+def delete_passkey_inline(
```
eniklas, line 165: Worth a conscious CSRF decision here. The previous flow used `DELETE` + `Content-Type: application/json`, which forces a CORS preflight and is effectively CSRF-safe. This new `POST .../delete` accepts a plain form submission with cookie auth (`current_user_api`), which is a "simple request" and can be triggered cross-origin. It's **mitigated** because deletion requires the account password in the body, which an attacker can't supply — so not exploitable today. But this looks like the app's first state-changing, cookie-authenticated form POST without a CSRF token, so it's worth noting in case the password gate is ever relaxed.

src/gamatrix/auth/routes.py
```diff
@@ -118,6 +134,61 @@ def list_passkeys(
...
+    repo: Repository = Depends(get_repo),
+):
+    passkey = _owned_passkey(credential_id, user, repo)
+    if not service.authenticate(repo, user["email"], password):
+        return templates.TemplateResponse(
+            request,
+            "passkey_delete_form.html.jinja",
+            {"passkey": passkey, "error": "Wrong password."},
+        )
+    if not repo.delete_passkey(credential_id, user["webauthn_user_id"]):
```
eniklas, line 179: Minor: `_owned_passkey(credential_id, ...)` on line 172 already verified ownership and raised 404, so this `repo.delete_passkey(...) -> False` 404 branch is now only reachable on a concurrent-deletion race. That's fine to keep as a guard, just flagging it's otherwise dead under normal flow.

src/gamatrix/templates/passkey_delete_form.html.jinja
```diff
@@ -0,0 +1,16 @@
+<li class="passkey-delete-confirm">
+    <strong>Delete {{ passkey.friendly_name }}?</strong>
+    <form hx-post="/auth/passkeys/{{ passkey.credential_id }}/delete"
+          hx-target="closest li" hx-swap="outerHTML">
+        <label>
+            Current password
+            <input type="password" name="password" required
```
eniklas, line 7: `autofocus` only fires when the browser parses the document on initial load — it is ignored on nodes inserted via an HTMX `outerHTML` swap, so the password field won't actually receive focus when the confirm form appears. If the auto-focus matters, add an `hx-on::after-swap` handler (or a small `htmx:afterSwap` listener / the `autofocus` extension) to call `.focus()` on the input.

eniklas (commented):
Nice cleanup — replacing the `window.prompt` flow with an inline HTMX form is a real UX/security improvement, and the cross-account ownership test is a good addition. A few things worth addressing before merge (none are blockers on the happy path):

1. **The old `DELETE /auth/passkeys/{credential_id}` endpoint is now orphaned.** This PR removes its only caller (the `.delete-passkey` handler in `passkeys.js`), so the JSON endpoint at routes.py:243 no longer has a frontend client and now duplicates the ownership/auth/delete logic that lives in `delete_passkey_inline` + `_owned_passkey`. The two paths can drift (e.g. the old one returns 403 on a bad password, the new one returns an inline 200). Recommend either deleting it or routing it through `_owned_passkey` so there's a single source of truth.
2. **CSRF surface change** — see inline note. Mitigated by the password gate, but worth a conscious decision.
3. **`autofocus` won't fire on the HTMX-swapped form** — minor UX nit, inline.

Details inline.

derek-keeler:
I'm going to go over the comments this week and fix them up. However, I think this was done too quickly and needs a bit more thought (the design phase was left completely up to codex, which seems to have missed the point on some things!).

Shouldn't be a blocker for the app though - I'm the only one using passkeys for now :)

derek-keeler:
Closing until I come back with a more well thought out plan.

PR #168, opened 2026-06-15
Show API token last-used time in the viewer's local time zone

eniklas:
## What

The API tokens management page rendered each token's `last_used_at` as the raw stored **UTC** date. This localizes it to the viewer's browser time zone, mirroring how the DB upload time was handled in #161.

## How

- `tokens_list.html.jinja`: render the UTC instant into a `data-last-used` attribute on a `.token-last-used` span, keeping the server-rendered date as a no-JS fallback.
- `tokens.html.jinja`: rewrite each span with `toLocaleString()`, which picks up the browser's locale and zone automatically. The list is fetched and re-fetched via HTMX (on load, and after create/revoke), so the rewrite runs on every `htmx:afterSwap` into `#token-list`.

`last_used_at` is stored via `now_iso()` as a UTC ISO string with a `+00:00` offset, so `new Date(iso)` parses it unambiguously — same as `db_updated_at`.

## Test

Added `test_token_list_exposes_last_used_for_client_localization`, asserting the rendered list carries the raw UTC timestamp in `data-last-used` once a token has been used (and not before). `just lint` and `pytest tests/test_tokens.py` pass.

### DIFF
--- a/src/gamatrix/templates/tokens.html.jinja
+++ b/src/gamatrix/templates/tokens.html.jinja
@@ -68,6 +68,23 @@
         htmx.ajax("GET", "/auth/tokens/list", { target: "#token-list" });
     }
 
+    // Last-used times are stored as UTC. Rewrite each one in the viewer's local
+    // time zone; toLocaleString() picks up the browser's locale and zone
+    // automatically. The list is fetched and re-fetched via HTMX, so run this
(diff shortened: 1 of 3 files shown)


derek-keeler (approved):
Yep simple I likey

PR #169, opened 2026-06-15
Personalize uploader scripts and rename the scheduled task

eniklas:
Two fixes to the unattended-upload setup handed out on the **API tokens** page.

## 1. Scripts pointed at a placeholder host

The downloadable uploader scripts (`upload-gamatrix.ps1` / `.sh`) shipped a generic `https://gamatrix.example.com` default, so anyone who ran one without overriding the base URL hit the wrong site.

The server already knows the deployment's real site (`app_base_url`), so the scripts are now served through `GET /auth/upload-gamatrix.ps1` and `GET /auth/upload-gamatrix.sh`, which read the script and substitute the placeholder for the configured base URL. The PowerShell setup snippet and the macOS/Linux download link reference those routes, so the copy a user gets points at the right host out of the box — no hand-editing.

(The routes are public, matching how the PowerShell `Invoke-WebRequest` fetches the script with no cookie.)

## 2. Scheduled-task name collided with v1

The setup snippet registered the Windows task as `'Gamatrix DB upload'` — the same name v1 used, which conflicts if a v1 task is still defined. Renamed to `'gamatrix v2 DB upload'`.

## Test

- Added `test_uploader_scripts_point_at_the_configured_site`: both served scripts return 200, contain the configured base URL, and no longer contain `gamatrix.example.com`.
- Extended the snippet test to assert the `/auth/upload-gamatrix.ps1` download URL and the `gamatrix v2 DB upload` task name.
- `just lint`, `just typecheck`, and `pytest tests/test_tokens.py` all pass.

### DIFF
--- a/src/gamatrix/auth/routes.py
+++ b/src/gamatrix/auth/routes.py
@@ -2,8 +2,15 @@
 
 from __future__ import annotations
 
+from pathlib import Path
+
 from fastapi import APIRouter, Depends, Form, HTTPException, Request, status
-from fastapi.responses import HTMLResponse, JSONResponse, RedirectResponse
+from fastapi.responses import (
+    HTMLResponse,
+    JSONResponse,
+    PlainTextResponse,
+    RedirectResponse,
+)
 from pydantic import BaseModel
 
 from gamatrix.auth import passkeys, service, tokens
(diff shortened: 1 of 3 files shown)


derek-keeler (approved):
Yep, 🚢 it

Issue #21, opened 2021-01-22
allow IP whitelist


eniklas:
Add IP whitelisting so this doesn't need to be done with iptables.

eniklas:
Done in https://github.com/eniklas/gamatrix-gog/pull/33.

PR #171, opened 2026-06-17
[1/3] Separate data and UX layers to make UX updates more independant

derek-keeler:
_Change [1 / 3] to get updated UX in place for Gamatrix: [Preview of the hotness is here](https://github.com/user-attachments/assets/c738d157-c53c-45a0-99d3-1dc1b4d92390)_

Necessary plumbing to build a truly independent data layer and UX layer. This will allow us to re-tool the UX and data layers separately from one another, allowing us to build more robust interfaces in the future (modern UX on the web, modes such as dark/light/color-blind/etc..., mobile views, even CLI or TUI if we feel adventurous!).

Breakdown of the high-impact changes in this feature branch.

### What

Separated the comparison domain from the web presentation layer so game comparison logic can be reused outside the current Jinja/HTMX flow. This introduces a typed comparison/query contract, a small web adapter layer, and a new `/api/games` JSON endpoint.

### How

`games/service.py` now owns a UX-neutral comparison model and refresh-planning logic, while `games/web.py` translates request/preferences state into that model and shapes results back into the existing page contract. `games/routes.py` was updated to use that split without changing the current authenticated template system, which keeps this PR focused on data/service boundaries and creates the seam the next PR uses to swap in templated authenticated UX.

### DIFF
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -3,8 +3,7 @@
 from __future__ import annotations
 
 from fastapi import APIRouter, Depends, Request
-from fastapi.responses import HTMLResponse
-from fastapi.responses import JSONResponse
+from fastapi.responses import HTMLResponse, JSONResponse
 
 from gamatrix.auth.dependencies import (
     current_user,
@@ -29,15 +28,6 @@ from gamatrix.templating import templates
 router = APIRouter(tags=["games"])
 
 
-def _parse_options(
-    request: Request,
-    user: dict,
-    repo: Repository | None = None,
-) -> WebCompareOptions:
-    """Build web options from saved preferences overlaid with query params."""
-    return web.parse_options(request, user, repo or get_repo())
-
-
 def _maybe_enrich(repo: Repository, opts: WebCompareOptions) -> str | None:
     """Find stale/pending games among the selected libraries and enqueue a job.
 
@@ -55,7 +45,7 @@ def games_page(
     user: dict = Depends(current_user),
     repo: Repository = Depends(get_repo),
 ):
-    opts = _parse_options(request, user, repo)
+    opts = web.parse_options(request, user, repo)
     job_id = _maybe_enrich(repo, opts)
     users = {str(u["user_id"]): u for u in repo.scan_users() if u.get("user_id")}
     result = service.compare(repo, opts.to_query())
@@ -84,7 +74,7 @@ def games_table(
     repo: Repository = Depends(get_repo),
 ):
     users = {str(u["user_id"]): u for u in repo.scan_users() if u.get("user_id")}
-    opts = _parse_options(request, user, repo)
+    opts = web.parse_options(request, user, repo)
     result = service.compare(repo, opts.to_query())
     caption = web.build_caption(users, opts, result)
     return templates.TemplateResponse(
@@ -107,7 +97,7 @@ def games_api(
     repo: Repository = Depends(get_repo),
 ):
     """Return the comparison dataset as JSON for headless consumers."""
-    opts = _parse_options(request, user, repo)
+    opts = web.parse_options(request, user, repo)
     query = opts.to_query()
     result = service.compare(repo, query)
     return JSONResponse(api.serialize_comparison(query, result))
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from fastapi import APIRouter, Depends, Request
 from fastapi.responses import HTMLResponse
+from fastapi.responses import JSONResponse
 
 from gamatrix.auth.dependencies import (
     current_user,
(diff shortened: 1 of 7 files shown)


src/gamatrix/games/routes.py
```diff
@@ -4,115 +4,49 @@
 
 from fastapi import APIRouter, Depends, Request
 from fastapi.responses import HTMLResponse
+from fastapi.responses import JSONResponse
```
eniklas: [NIT] could be collapsed into the above import.

src/gamatrix/games/service.py
```diff
@@ -139,32 +189,32 @@ def _multiplayer(max_players: int, game_modes: list[int]) -> bool:
...
     """
-    grouped: dict[tuple[str, frozenset], dict] = {}
+    grouped: dict[tuple[str, frozenset[str]], ComparisonItem] = {}
     for game in games:
-        key = (game["slug"], frozenset(game["owners"]))
+        key = (game.slug, frozenset(game.owners))
         if key not in grouped:
-            grouped[key] = {**game, "platforms": list(game["platforms"])}
+            grouped[key] = ComparisonItem(
+                **{**game.to_dict(), "platforms": list(game.platforms)}
```
eniklas: This clones the first-seen item by round-tripping through `asdict` and then re-overriding `platforms`. Since `to_dict()`/`asdict` already deep-copies every list, the `"platforms": list(game.platforms)` override is redundant. `dataclasses.replace` reads cleaner and skips the recursive asdict per first-seen game:

```python
from dataclasses import replace
...
grouped[key] = replace(game, platforms=list(game.platforms))
```

`replace` shares the `installed`/`owners` lists with the source item, but those are only ever rebound below (`existing.installed = sorted(...)`), never mutated in place, so it's safe — only `platforms` gets `.append()`ed, and that's the one we copy.

src/gamatrix/games/service.py
```diff
@@ -1,58 +1,111 @@
...
+    items: list[ComparisonItem]
     excluded_user_ids: list[str]
     total: int
 
 
-def compare(repo: Repository, opts: CompareOptions) -> CompareResult:
+@dataclass
+class RefreshAdvice:
+    job_id: str | None
+    stale_release_keys: list[str] = field(default_factory=list)
```
eniklas, line 96: `stale_release_keys`, `reused_active_job`, and `created_job` are populated here but no caller reads them — `routes._maybe_enrich` only uses `advice.job_id`. If these are seams for the later PRs in the series that's fine; otherwise they're dead fields worth trimming until something consumes them.

src/gamatrix/games/routes.py
```diff
@@ -4,115 +4,49 @@
...
-        sort=qp.get("sort", "title"),
-        direction=qp.get("dir", "asc"),
-    )
+def _parse_options(
+    request: Request,
+    user: dict,
+    repo: Repository | None = None,
+) -> WebCompareOptions:
+    """Build web options from saved preferences overlaid with query params."""
+    return web.parse_options(request, user, repo or get_repo())
```
eniklas: Every production caller now passes `repo` explicitly, so the `repo or get_repo()` fallback only fires for the `tests/test_routes.py::_opts` helper. Minor, but you could drop the fallback and have the test pass a repo (it already constructs a fake one), which would let `_parse_options` be a straight pass-through to `web.parse_options`.

src/gamatrix/games/routes.py
```diff
@@ -121,18 +55,18 @@ def games_page(
     user: dict = Depends(current_user),
     repo: Repository = Depends(get_repo),
 ):
-    opts = _parse_options(request, user)
+    opts = _parse_options(request, user, repo)
     job_id = _maybe_enrich(repo, opts)
-    result = service.compare(repo, opts)
-    caption = service.build_caption(repo, opts, result)
     users = {str(u["user_id"]): u for u in repo.scan_users() if u.get("user_id")}
```
eniklas, line 50: Small efficiency note (pre-existing, and this PR already improved it by dropping the scan in `build_caption`): the route scans users here and `service.compare` scans them again internally. Since `compare` needs the full user set anyway for exclusive mode, you could have it return the `users` dict (or accept it) and avoid the second `scan_users()` per request. Not blocking.

eniklas (commented):
Nice refactor — the service/web split is clean and the template contract is preserved (`present_games` still emits the same dict keys, and `WebCompareOptions` keeps the `.all_games`/`.sort` shape the templates expect). Verified `test_games.py` + `test_routes.py` pass on the branch. A few small, non-blocking observations inline.

src/gamatrix/games/service.py
derek-keeler: I am not 100% certain if they are used until later. I've made a note to myself to follow up on this once the next PR lands.

See https://github.com/derek-keeler/gamatrix/issues/21

src/gamatrix/games/service.py
derek-keeler: TIL!

Added some tests around this functionality to ensure I preserved contracts/functionality and then ran those tests before and after to ensure that I didn't break anything. Good one.

src/gamatrix/games/routes.py
derek-keeler: Yep, this was a nonsense step with the current codebase. Removed it in favour of the web.parse_options call directly.

src/gamatrix/games/routes.py
derek-keeler: I agree that this is doing redundant things. Looking into it deeper I think we can tighten up the contract a bit - right now the Repository type is defining 'scan_users' very loosely and that is propagated via the ComparisonRepository's Protocol declaration. I feel this is worthy of a bit more of a deeper refactor and would like to leave it out of the changeset I'm proposing here.

A revamp of the data layer will be far simpler to handle once this PR lands :) so we have that to look forward to.

eniklas (approved):

PR #172, opened 2026-06-17
Introduce new template framework to Gamatrix with user-selectable visual modes

### LINKED ISSUE #92: Investigate UX enhancements

derek-keeler:
_Change [2 / 3] to get updated UX in place for Gamatrix: [Preview of the hotness is here](https://github.com/user-attachments/assets/c738d157-c53c-45a0-99d3-1dc1b4d92390)_

Builds on  #171 - fully templatizes the current 'default' UX, showing the new independent UX layer is functional and fully independent.

### What

Moves the current authenticated UX into a `default` template set and adds per-user display mode preferences, while keeping the existing user-facing flow intact. (Authenticated UX means "all UX except the parts visible prior to login"). This makes the authenticated UX themeable without yet introducing a second visual template.

### How

Added an authenticated template wrapper, nested `templates/default/` and matching static assets, plus preference persistence/validation for `display_mode`. The routes now render authenticated pages through that wrapper, so the next PR can add `modern` as a deployment-selected template instead of rewriting route behavior again.

In support of #92 

NOTE: There will be no/slight visual changes when this is merged. The point was to move the existing UX into a template (called 'default') that will support the new system and also provide the new modes (dark/light/color-blind/high-contrast). It does not provide implementations for the modes - it just supports them in the preferences page (you can switch between them without causing errors, but no visual change will be apparent). The logic in doing this was that we are likely going to remove that template with something else - the `modern` template in the next PR or something even more basic.


derek-keeler:
Note that there was some concern over how this set of changes will affect the UX treatment for resolving IGDB enrichment update statuse from the first PR. I am tracking that potential regression here: derek-keeler/gamatrix#21 - will fully resolve that in the last PR in this series (and will ensure no regressions).

src/gamatrix/auth/routes.py
```diff
@@ -19,7 +19,7 @@
 from gamatrix.config import get_settings
 from gamatrix.constants import API_TOKEN_NAME_MAX_LENGTH
 from gamatrix.storage.dynamo import Repository
-from gamatrix.templating import templates
+from gamatrix.templating import authenticated_template, templates
```
eniklas, line 22: `templates` is still imported here because the **API-token management pages remain on the old root environment** — `token_management` and `list_tokens` (further down, unchanged in this diff) still call `templates.TemplateResponse(... "tokens.html.jinja"/"tokens_list.html.jinja")`. Those are authenticated UX (gated by `current_user`) and the preferences page links straight to `/auth/tokens`, but they:

- render through root `base.html.jinja` + `/static/style.css` instead of `default/base.html.jinja` + `default/style.css`, and
- never get the `data-mode` attribute, so they won't honor the user's `display_mode`.

They're also absent from `AUTHENTICATED_TEMPLATE_NAMES`, so the "complete template contract" test doesn't catch the gap. Right now there's no visible difference, but once PR 3 swaps `default/` for `modern/`, the token pages will be stranded on the original look. Recommend moving `tokens.html.jinja` / `tokens_list.html.jinja` into `default/` and routing them through `authenticated_template` too (the `forgot_password`/`reset_password`/`login` pages are correctly left on root since they're pre-auth).

src/gamatrix/templating.py
```diff
@@ -1,18 +1,63 @@
...
+from starlette.responses import HTMLResponse
 
 from gamatrix import __version__
 from gamatrix.constants import PLATFORMS
+from gamatrix.games.preferences import merge_preferences
 from gamatrix.helpers import pic_url
 
 TEMPLATES_DIR = Path(__file__).parent / "templates"
-
+AUTHENTICATED_TEMPLATE_NAMES = (
```
eniklas, line 17: This contract list omits `tokens.html.jinja` and `tokens_list.html.jinja`, which are part of the authenticated UX (see the auth/routes comment). If they get migrated into `default/`, add them here so the contract test guards them as well.

Separately: the seven original root authenticated templates (`games.html.jinja`, `games_table.html.jinja`, `job_status.html.jinja`, `passkeys.html.jinja`, `passkeys_list.html.jinja`, `preferences.html.jinja`, `upload.html.jinja`) are now unreferenced — every route renders the `default/` copies. They're dead duplicates and a drift hazard (an edit could land on the wrong copy). Worth deleting them in this PR unless something still depends on them.

src/gamatrix/templating.py
```diff
@@ -1,18 +1,63 @@
...
+    request,
+    name: str,
+    context: dict,
+    status_code: int = 200,
+    headers: dict[str, str] | None = None,
+    media_type: str | None = None,
+    background: BackgroundTask | None = None,
+) -> HTMLResponse:
+    """Render an authenticated page or fragment with the account's saved mode."""
+    user = context.get("user") or {}
```
eniklas, line 55: Minor: for the fragment templates that don't extend `base.html.jinja` (`games_table`, `job_status`, `passkeys_list`), the injected `display_mode` is unused — harmless, just noting the `"user": user` additions to those route contexts in this PR aren't doing anything for theming. Also the annotated return type is `HTMLResponse` but `TemplateResponse` returns a starlette `_TemplateResponse`; fine functionally, just slightly imprecise.

pyproject.toml
```diff
@@ -78,8 +78,8 @@ fallback_version = "0.0.0"
 where = ["src"]
 
 [tool.setuptools.package-data]
-"gamatrix.templates" = ["*.jinja"]
-"gamatrix.static" = ["*.png", "*.jpg", "*.css", "*.js"]
+"gamatrix.templates" = ["*.jinja", "**/*.jinja"]
+"gamatrix.static" = ["*.png", "*.jpg", "*.css", "*.js", "**/*.css"]
```
eniklas: `**/*.css` packages the nested `default/style.css` correctly, but nested PNG/JPG/JS under `static/` still aren't globbed recursively (only top-level `*.png`/`*.jpg`/`*.js`). Fine for this PR since `default/` only ships CSS, but if a future template set adds nested images or JS they'd silently be left out of the wheel. Consider `**/*` patterns for those too, or a comment noting the constraint.

eniklas (commented):
Solid step 2 — the second Jinja environment rooted at `templates/default/` plus the `authenticated_template()` wrapper is a clean seam, and the `display_mode` validation (round-tripping through `merge_preferences` and the `system`→`None` mapping) is well covered by the new tests. I checked out the branch and `test_preferences.py` + `test_templates.py` pass (20 passed).

My main concern is that the migration to `default/` is incomplete: the API-token management pages are authenticated UX but were left on the old root environment, and the seven original root authenticated templates are now orphaned duplicates. Details inline — neither is visually apparent today (default renders identically), but both undercut the "swap the template set in PR 3" goal this PR is setting up.

src/gamatrix/templates/default/base.html.jinja
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
```
eniklas, line 1: I'm sure there's a good reason for this, but for my own education: curious why the Jinja files are under a theme subdir; I expected the CSS would handle all the theming. Does this mean these will be duplicated for each theme going forward?

derek-keeler: Yep - duplicated (well, sorta - _replicated and updated_ would be the way I'd describe it). My reasoning for this is so we can specify different themes at deploy time.

For example, in the next PR you will be able to chose between 'default' and 'modern'.

In the future, if you wanted to do a Bob & Doug theme, you could keep the 'prod' deploying 'default' or 'modern', but be iterating on 'bobanddoug' locally, checking in things and tweaking things until you get it to the point where you update and redeploy with that theme... basically you can have multiple themes ready to go at any time.

src/gamatrix/auth/routes.py
derek-keeler: Good catch. Will fix this up - in my 3rd PR I see these files are still excluded from the 'UX' contract.

src/gamatrix/templates/default/base.html.jinja
eniklas: I see, I thought themes were selectable at runtime per user.

src/gamatrix/templating.py
derek-keeler: These were left out - probably sync issues when things were being done. Thanks for catching.

Fixed this and also removed orphaned original UX jinja.

src/gamatrix/templates/default/base.html.jinja
derek-keeler: Themes are deploy-time. _Modes_ are selectable at runtime.

I made this (potentially mistaken) decision based on templating opening up some fun user experience and security issues. (Change to template X, hit this endpoint-that-only-exists-in-template-Y).

src/gamatrix/auth/routes.py
derek-keeler: This was fixed alongside the `src/gamatrix/templating.py` issue (below).

src/gamatrix/templating.py
derek-keeler: (1) Keeping `"user": user` on the fragment routes is harmless and _slightly_ defensive:
- it keeps the route call shape consistent across authenticated renders
- it avoids special-casing fragments in the route layer
- if one of those fragments later grows a base-extending wrapper or starts using user-driven presentation state, the route already provides the needed context.

The cleanup was done however, as we don't want to add noise or imprecision into the app at this early stage

pyproject.toml
derek-keeler: Package specifier in pyproject.toml allows for nested static PNG/JPG/JS

src/gamatrix/templates/default/base.html.jinja
eniklas: Ah, so modes are strictly cosmetic while themes can alter functionality?

eniklas (approved):

src/gamatrix/templates/default/base.html.jinja
derek-keeler: Yep!

True functionality is limited by the API of course, but the template can alter how it makes use of the API.

Generally speaking, templates will be thematic (think modern, tron, all-comic-sans, retro-arcade, lipstick on pig kind of thing), but they could alter the views available to the user (mobile, watch).

The 'modes' are like the colour scheme, kinda like the modes in a desktop manager.

PR #173, opened 2026-06-19
[3 / 3] Add the first non-OG "modern" template

derek-keeler:
_Change [3 / 3] to get updated UX in place for Gamatrix: [Preview of the hotness is here](https://github.com/user-attachments/assets/c738d157-c53c-45a0-99d3-1dc1b4d92390)_

Add a new 'modern' template - final state with a fully-template-able UX for Gamatrix. All modes are supported and the entire 

### What

Adds a second authenticated template, `modern`, and makes the deployment choose between `default` and `modern`. This keeps the data and display-mode work from earlier PRs intact while introducing the new visual layer as an additive change.

### How

Added `templates/modern/` and its stylesheet, then generalized the authenticated templating setup to validate and load the configured UX template. Runtime and CDK config now pass `UX_TEMPLATE` through deployment, with tests covering template selection and config validation.

Change your template by setting `UX_TEMPLATE=modern` in your .env file prior to deployment. Values are: `default` or `modern`.

### DIFF
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -80,6 +80,7 @@ def games_table(
         request,
         "games_table.html.jinja",
         {
+            "user": user,
             "users": users,
             "games": web.present_games(result, opts),
             "caption": caption,
@@ -118,7 +119,7 @@ def job_status(
     return authenticated_fragment(
         request,
         "job_status.html.jinja",
-        {"job": job, "job_id": job_id, "done": done},
+        {"user": user, "job": job, "job_id": job_id, "done": done},
     )
 
 
@@ -144,6 +145,7 @@ def refresh_missing(
         request,
         "job_status.html.jinja",
         {
+            "user": admin,
             "job": repo.get_job(job_id) if job_id else None,
             "job_id": job_id,
             "done": job_id is None,
@@ -169,6 +171,7 @@ def refresh_all(
         request,
         "job_status.html.jinja",
         {
+            "user": admin,
             "job": repo.get_job(job_id) if job_id else None,
             "job_id": job_id,
             "done": job_id is None,
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -81,7 +81,6 @@ def games_table(
         request,
         "games_table.html.jinja",
         {
-            "user": user,
             "users": users,
             "games": web.present_games(result, opts),
             "caption": caption,
@@ -120,7 +119,7 @@ def job_status(
     return authenticated_fragment(
         request,
         "job_status.html.jinja",
-        {"user": user, "job": job, "job_id": job_id, "done": done},
+        {"job": job, "job_id": job_id, "done": done},
     )
 
 
@@ -146,7 +145,6 @@ def refresh_missing(
         request,
         "job_status.html.jinja",
         {
-            "user": admin,
             "job": repo.get_job(job_id) if job_id else None,
             "job_id": job_id,
             "done": job_id is None,
@@ -172,7 +170,6 @@ def refresh_all(
         request,
         "job_status.html.jinja",
         {
-            "user": admin,
             "job": repo.get_job(job_id) if job_id else None,
             "job_id": job_id,
             "done": job_id is None,
--- a/src/gamatrix/games/routes.py
+++ b/src/gamatrix/games/routes.py
@@ -28,17 +28,6 @@ from gamatrix.templating import authenticated_fragment, authenticated_template
 router = APIRouter(tags=["games"])
 
 
-def _maybe_enrich(repo: Repository, opts: WebCompareOptions) -> str | None:
-    """Find stale/pending games among the selected libraries and enqueue a job.
-
-    If a job is already pending or running, return its id rather than
-    creating a duplicate — each /games page load would otherwise queue a
-    new batch of the same games, exhausting Lambda concurrency.
-    """
-    advice = service.ensure_enrichment_job(repo, get_queue(), opts.to_query())
-    return advice.job_id
-
-
 @router.get("/games", response_class=HTMLResponse)
 def games_page(
     request: Request,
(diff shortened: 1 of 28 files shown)


src/gamatrix/templates/modern/base.html.jinja
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>{% block title %}gamatrix{% endblock %}</title>
+    {# Version query busts the browser cache on each deploy so CSS changes
+       (e.g. the sticky results header) aren't masked by a stale stylesheet. #}
+    <link rel="stylesheet" href="/static/templates/modern/style.css?v={{ version }}">
```
eniklas, line 9: This stylesheet href is **the only real difference between the entire `modern` and `default` template sets**. I diffed all ten: `games`, `games_table`, `job_status`, `passkeys`, `passkeys_list`, `preferences`, `tokens`, `tokens_list`, `upload` are byte-identical between the two sets, and `base.html.jinja` differs only in this one line (`default/` vs `modern/` in the path). So ~780 lines of Jinja are duplicated to vary a single URL.

That's a real maintenance hazard: every future fix to the games table, preferences form, token UI, etc. now has to be made in two (eventually N) places or the sets silently diverge — and the contract test only checks that each template *renders*, not that they stay behaviorally in sync.

Since the modes are already entirely CSS-variable driven, the visual layer is fully carried by `style.css` alone. Consider keeping one shared authenticated template set and parameterizing just the stylesheet, e.g. inject the selected name as a Jinja global and render `href="/static/templates/{{ ux_template }}/style.css?v={{ version }}"`. That collapses the duplication to one template set + one stylesheet per theme. If full per-template divergence is genuinely expected later (different markup, not just colors), it'd help to say so here — right now the duplication buys nothing the CSS doesn't already provide.

src/gamatrix/templating.py
```diff
@@ -26,6 +27,7 @@
     "tokens_list.html.jinja",
     "upload.html.jinja",
 )
+UX_TEMPLATES = ("default", "modern")
```
eniklas: The set of valid template names now lives in three places: this tuple, `Settings.ux_template`'s `Literal["default", "modern"]` in `config.py`, and `UX_TEMPLATES` in `infrastructure/cdk/config.py`. Adding a third theme means editing all three, and they can drift (e.g. CDK accepts a name the runtime Literal rejects). The two in-package copies can share a single source — derive this tuple from the Literal via `typing.get_args(...)` (or define the tuple once and build the Literal from it). The CDK copy is necessarily separate, but a comment cross-referencing it would help.

src/gamatrix/games/routes.py
```diff
@@ -81,6 +81,7 @@ def games_table(
         request,
         "games_table.html.jinja",
         {
+            "user": user,
```
eniklas: This `"user": user` (and the `"user": admin` entries added to the two refresh routes + the `job_status` route) is dead context: `games_table`/`job_status` are rendered via `authenticated_fragment`, which — unlike `authenticated_template` — intentionally doesn't read `user` to compute `display_mode`, and the fragment templates only reference `users`, never `user`. So these additions have no effect. Either drop them, or if the intent was for fragments to also pick up the saved display mode, that belongs in `authenticated_fragment` (though fragments have no `<body>`, so the `data-mode` attribute wouldn't apply anyway). Worth clarifying which you intended. (The `"user": user` on line 57 in `games_page` is fine — that one goes through `authenticated_template`.)

src/gamatrix/templating.py
```diff
@@ -37,9 +39,16 @@ def _configure(environment: Jinja2Templates) -> Jinja2Templates:
...
+
+
+def build_authenticated_templates(template_name: str) -> Jinja2Templates:
+    """Build an environment for a validated deployment-selected UX template."""
+    if template_name not in UX_TEMPLATES:
+        raise ValueError(f"Unknown UX template: {template_name}")
+    return _configure(Jinja2Templates(directory=str(TEMPLATES_DIR / template_name)))
+
+
+authenticated_templates = build_authenticated_templates(get_settings().ux_template)
```
eniklas: Minor: this evaluates `get_settings()` at import time, so importing `gamatrix.templating` now requires settings to load successfully — a new import-time coupling this module didn't have before. It's fine in practice (settings have defaults and `get_settings` is cached), just worth being aware that an env/validation error would now surface as an import failure here rather than at first request.

eniklas (commented):
Closes out the series nicely — the deploy-time `UX_TEMPLATE` selection is wired cleanly from CDK config → Lambda env → `Settings` → `build_authenticated_templates`, with good validation at both the CDK and runtime layers, and the parametrized template/stylesheet tests are a solid contract. I checked out the branch and the affected suites pass (31 passed). Nice that the token-template gap I flagged on #172 is now closed (both sets ship `tokens.html.jinja`).

One structural concern dominates: the `modern` set is almost entirely a byte-for-byte copy of `default`, which sets up a long-term drift hazard. Details + a couple of smaller notes inline.

src/gamatrix/templates/modern/base.html.jinja
derek-keeler: This is _by design_ and it shouldn't be fixed.

This was the first true template UX that Gamatrix will see. It was intended to be an in-place replacement, not a departure from, the OG UX. Once this is merged we can consider deleting the 'default' or replacing it with the 'modern' template, and templates that will come after will be the ones that will attempt true departures.

We also are keen to avoid _coupling_ anything used between disparate templates. As we progress from here, we can learn from the patterns we see as we add more templates and coalesce what makes sense.

derek-keeler:
Will investigate/fix issues in the morning.

src/gamatrix/templating.py
derek-keeler: Removed the need for hard-coded reference to the valid templates from the code completely, and available templates are stored in the production package file under the templates/ folder. Valid templates are demarcated with a 'valid_template' file in the root folder of all templates that are usable in production, so we can differentiate between valid templates and ones that are not yet ready for production. Adding a new theme will simply mean adding a folder under the `templates/` folder, making it work, then adding the empty `valid_template` file to its root.

src/gamatrix/games/routes.py
derek-keeler: Will remove the unused "user" information. If we need it later on then we can re-add it then (no evidence we will at this time).

src/gamatrix/templating.py
derek-keeler: I've put this in the app setup now, if the settings cause an exception then that will be in the application code and not at template load time. (Easier to catch in a debugger or deduce from a callstack).

derek-keeler:
Also in this change:

- Fix for regression: database upload time in user's own timezone in hovertext on main page.
- Fix for derek-keeler/gamatrix#21.

eniklas (approved):

PR #174, opened 2026-06-22
Make modern "Dark" mode use the pre-template (2.3.x) palette

eniklas:
### What

Repoints the **modern** template's existing **Dark** mode at the color scheme Gamatrix used before the templating framework was introduced (e.g. `2.3.3`): a dark page with the original light-blue results table, green/red ownership cells, and light-cyan accents.

No new mode — selecting **Dark** in Preferences (or having an OS dark preference) now gives the v1 look.

### How

Single file: `static/templates/modern/style.css`.

- `body[data-mode="dark"]` palette swapped from the previous modern dark (dark rows, light text) to the v1 palette (dark page, **light** rows, dark cell text).
- Because v1 inverts modern's text/row relationship, one extra rule flips the **table body** text back to dark so it stays readable on the light-blue rows; the page chrome keeps the light body text.
- The `@media (prefers-color-scheme: dark)` system default is updated to the same palette, keeping the two dark triggers in sync as the stylesheet already did. Its text-flip is guarded with `body:not([data-mode])` so it only applies to the system case and never leaks into an explicit non-dark mode (e.g. high-contrast) while the OS is dark.

### Notes

- Per your direction, this replaces the earlier approach of adding a separate `v1-dark` mode — `DISPLAY_MODES`, the preference dropdowns, and the `default` stylesheet are all back to their master state; the diff is now just the modern stylesheet.
- I updated **both** the explicit-Dark block and the system-dark `@media` block so "Dark" looks the same however it's triggered (they were already kept identical). If you only wanted the explicit `data-mode="dark"` choice changed and the system default left as the old modern dark, say the word and I'll narrow it.
- Faithful-to-v1 quirk carried over: a game with a manual `url` renders a light-cyan title link on a light-blue row (low contrast), exactly as v1 did.

### Testing

- `uv run pytest` — 155 passed (mode-contract + preference tests included).
- Verified `base.html.jinja` renders `<body data-mode="dark">` and the modern stylesheet stays well under the 20 KB budget.

### DIFF
--- a/src/gamatrix/static/templates/modern/style.css
+++ b/src/gamatrix/static/templates/modern/style.css
@@ -18,37 +18,46 @@ body[data-mode="light"] {
 
 @media (prefers-color-scheme: dark) {
   :root {
-    --page: #111b20;
-    --surface: #19272d;
-    --surface-alt: #22333a;
-    --text: #e4edf0;
-    --muted: #aabcc3;
-    --border: #49616a;
-    --accent: #6ed7df;
-    --accent-text: #102329;
-    --row-a: #19272d;
-    --row-b: #203239;
-    --owned: #28573e;
-    --unowned: #673c3c;
-    --warning: #ffd166;
+    --page: rgb(50, 50, 50);
+    --surface: #3c3c3c;
+    --surface-alt: #4a4a4a;
+    --text: lightgray;
+    --muted: #aaaaaa;
+    --border: #555555;
+    --accent: rgb(124, 207, 228);
+    --accent-text: #102027;
+    --row-a: rgb(180, 227, 239);
+    --row-b: rgb(201, 235, 243);
+    --owned: rgb(182, 215, 168);
+    --unowned: rgb(234, 153, 153);
+    --warning: grey;
   }
+  /* Only flip the table body text when no explicit mode is chosen (system
+     follows the OS). An explicit choice is handled by its own block below,
+     so this must not leak into e.g. high-contrast while the OS is dark. */
+  body:not([data-mode]) table.results tbody tr { color: var(--accent-text); }
 }
 
+/* Dark mode uses the pre-template (2.3.x) palette: a dark page chrome with the
+   original light-blue results table. v1 paired a dark page with *light* rows and
+   dark cell text, so the table body text is flipped back to dark below while the
+   page chrome keeps the light --text. */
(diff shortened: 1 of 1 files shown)


eniklas:
@derek-keeler I like the `modern` layout tweaks but miss the original color scheme.

derek-keeler:
No need to make a new mode - just overwrite the 'dark' mode in the modern template.

The mode is intended to be dark, and we can change the mode however the f we want when we host the thing ourselves. :)
- I put _zero_ effort into the modern template, like, _zero_. I didn't even _look_ at it when I got the clankers to make it up for me.
- I couldn't visual design myself out of a paper bag.
- I just wanted to be able to support different modes.
- I totally could have said "make the 'dark' mode _precisely, visually_ the same as the OG UX. I didn't.

We should probably talk about naming... "template" -> "mode" probably isn't cutting it. Probably more like "Theme" and "View Style"?

eniklas:
I find the concept of the templates a bit confusing, even after re-reading some of your past PRs explaining it. For instance:

> Generally speaking, templates will be thematic (think modern, tron, all-comic-sans, retro-arcade, lipstick on pig kind of thing), but they could alter the views available to the user (mobile, watch).

That sounds like you could use one template for mobile, and another for watches, but the template is a deploy-time option so you can only offer one at a time. Since we'll only be hosting one instance of this I don't really see the use (as opposed to modes, which are selectable per-user at runtime and make perfect sense to me).

As for the terms, maybe "template" -> "UX"? "Mode" feels right to me, e.g. "dark mode" and "color blind mode."

eniklas:
Updated to replace the existing `dark` mode with the original color scheme.

derek-keeler:
> I find the concept of the templates a bit confusing, even after re-reading some of your past PRs explaining it...
> ...
> That sounds like you could use one template for mobile, and another for watches, but the template is a deploy-time option so you can only offer one at a time. Since we'll only be hosting one instance of this I don't really see the use (as opposed to modes, which are selectable per-user at runtime and make perfect sense to me).
> 
> As for the terms, maybe "template" -> "UX"? "Mode" feels right to me, e.g. "dark mode" and "color blind mode."

That's fair. Let's chat about it in person one day and we can decide where to take it.

The UX and data layer are separate. That was the W for me.

derek-keeler (approved):
🚢 it!

PR #175, opened 2026-06-25
Speed up comparison sort and filtering

eniklas:
## Problem

Sorting a column with all ~2560 titles shown took **~4s**. Each column click fired a full `/games/table` HTMX request that re-ran the entire `compare()` pipeline against DynamoDB — user scan, per-user library queries, ~26 sequential `BatchGetItem` calls, a metadata scan — and re-rendered every row, just to reorder data the browser already had. Filter changes paid the same I/O cost.

## Changes

**1. Client-side sort (sort no longer touches the server)**
- `static/games_sort.js`: one delegated click listener reorders the `<tr>` rows already in the DOM via `data-sort-*` attributes. Survives HTMX swaps, so nothing to re-bind after a filter change.
- Both `games_table.html.jinja` templates: sort headers dropped `hx-get` and gained `class="sortable"` + `data-sort-col/-type` + `aria-sort`; rows carry `data-sort-title/players/rating/installed` (matching the old `SORT_KEYS`). The server still renders the initial order — pure progressive enhancement.

**2. 60s in-process read cache on the singleton `Repository`**
- New `Settings.read_cache_ttl_seconds` (default 60).
- TTL cache wrapping `scan_users`, `get_all_metadata`, per-user `get_user_library`, and a new `get_all_games_map` (one cached full-table scan replacing the per-request `BatchGetItem` chain in `compare()`).
- Every write path invalidates its key in-process (`put_game`, user writes, `replace_user_library`/`clear_user_library`, metadata writes).
- Filter changes become "cached reads + in-memory merge/filter" instead of re-scanning DynamoDB.

**3. flake8 cleanup**
- Removes three pre-existing unused imports (`routes.py`, `templating.py`, `test_routes.py`); `UX_TEMPLATES` is now imported in the test from its canonical module `ux_templates`.

## Tradeoff

Writes from the upload/enrich Lambdas only appear in the web process after its cache expires — up to the TTL (60s). Tune via `read_cache_ttl_seconds`.

## Testing

- `just check` green: black, flake8, mypy all clean.
- **161 tests pass** (155 existing + 6 new in `tests/test_repository_cache.py` covering cache hits, per-write invalidation, per-user isolation, and TTL=0 disabling the cache).

### DIFF
--- a/src/gamatrix/config.py
+++ b/src/gamatrix/config.py
@@ -74,6 +74,13 @@ class Settings(BaseSettings):
     igdb_stale_days: int = 30
     ux_template: str = DEFAULT_UX_TEMPLATE
 
+    # How long the Repository serves the comparison read-model (users, games,
+    # libraries, metadata overrides) from its in-process cache before re-reading
+    # from DynamoDB. Lets repeated filter changes reuse one set of reads. Writes
+    # invalidate the relevant cache entry in-process, so the only staleness is
+    # across separate processes (e.g. an upload/enrich Lambda) until expiry.
+    read_cache_ttl_seconds: float = 60.0
+
     # SSM parameter names for the title filter lists (AWS only). Locally these
     # are seeded into DynamoDB config and read from there.
     hidden_games_param: str = "/gamatrix/hidden-games"
--- a/src/gamatrix/static/games_sort.js
+++ b/src/gamatrix/static/games_sort.js
@@ -1,0 +1,93 @@
+// Client-side column sorting for the games results table.
+//
+// The full result set is already in the DOM, so sorting a column is just a
+// reorder of existing <tr> rows — no server round-trip. Each row carries the
+// sort keys as data-sort-<col> attributes; each sortable header carries
+// data-sort-col and data-sort-type ("text" or "num"). The server still renders
+// the initial order, so this is pure progressive enhancement.
+//
+// A single delegated listener on document survives HTMX swaps that replace the
+// table when filters change, so there is nothing to re-bind after a swap.
+(function () {
+    "use strict";
+
+    function cellValue(row, col, type) {
+        const raw = row.getAttribute("data-sort-" + col) || "";
+        if (type === "num") {
+            const n = parseFloat(raw);
+            return isNaN(n) ? 0 : n;
+        }
+        return raw.toLowerCase();
+    }
+
+    function nextDirection(th) {
+        // First click on a column sorts ascending; clicking the active
+        // ascending column flips to descending. Mirrors the old server logic.
+        return th.getAttribute("aria-sort") === "ascending"
+            ? "descending"
+            : "ascending";
+    }
+
+    function updateIndicators(thead, activeTh, direction) {
+        for (const th of thead.querySelectorAll("th.sortable")) {
+            const arrow = th.querySelector(".sort-arrow");
+            if (th === activeTh) {
+                th.setAttribute("aria-sort", direction);
+                if (arrow) {
+                    arrow.textContent = direction === "ascending" ? "▲" : "▼";
+                }
+            } else {
+                th.removeAttribute("aria-sort");
+                if (arrow) {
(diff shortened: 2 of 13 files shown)


src/gamatrix/config.py
```diff
@@ -74,6 +74,13 @@ class Settings(BaseSettings):
     igdb_stale_days: int = 30
     ux_template: str = DEFAULT_UX_TEMPLATE
 
+    # How long the Repository serves the comparison read-model (users, games,
+    # libraries, metadata overrides) from its in-process cache before re-reading
+    # from DynamoDB. Lets repeated filter changes reuse one set of reads. Writes
+    # invalidate the relevant cache entry in-process, so the only staleness is
+    # across separate processes (e.g. an upload/enrich Lambda) until expiry.
+    read_cache_ttl_seconds: float = 60.0
```
derek-keeler, line 82: Any reason this cannot be ~24 hours?

derek-keeler: It could also be split between the various cache ttls if that helps...

src/gamatrix/static/games_sort.js
```diff
@@ -0,0 +1,93 @@
...
+//
+// The full result set is already in the DOM, so sorting a column is just a
+// reorder of existing <tr> rows — no server round-trip. Each row carries the
+// sort keys as data-sort-<col> attributes; each sortable header carries
+// data-sort-col and data-sort-type ("text" or "num"). The server still renders
+// the initial order, so this is pure progressive enhancement.
+//
+// A single delegated listener on document survives HTMX swaps that replace the
+// table when filters change, so there is nothing to re-bind after a swap.
+(function () {
```
derek-keeler, line 11: Love this change. Should make things nice and snappy (or at least snappi-er).

No tests for this though. That something we can fix up? (I am asking for a friend that _will_ break this someday, but keeping them anonymous to ensure they are protected).

derek-keeler (approved):
Looks good. Great upgrade. Some non-blocking comments that can be fully ignored/pushed to a separate PR if you like.

src/gamatrix/config.py
eniklas: It can technically, but the TTL is also the cross-process staleness window. The web Lambda invalidates its own cache keys on writes, but an upload or enrichment Lambda writes straight to DynamoDB with no way to signal the web process — so the TTL is the worst-case delay before a freshly uploaded DB or a completed enrichment shows up in the comparison view. At 24h someone could upload their library and not see it for a day. 60s keeps that window small while still collapsing a burst of filter/sort clicks into a single set of reads.

The split is the better long-term knob though — the games table and metadata overrides change rarely and could tolerate a much longer TTL, while user/library reads (which reflect uploads) want to stay short. Happy to do per-entry TTLs in a follow-up if we want to push the games scan way up.

src/gamatrix/static/games_sort.js
eniklas: Fair catch. There's no JS test harness in the repo today — it's pure Python, no node/jest/vitest — so this was the gap. Two ways to close it in a follow-up: (a) add a minimal node + jsdom test covering the comparator and row reorder, which means introducing a small JS toolchain; or (b) extract the pure sort/compare logic out of the click handler so it's unit-testable without a DOM, then smoke-test that. I'd lean (b) to avoid standing up node tooling for one file. Can do either in a separate PR.

Your friend's identity remains fully protected.

src/gamatrix/config.py
derek-keeler: Yeah - lets leave the split TTL for now and put this in now. If see any major spikes/slowdowns we can investigate later.

src/gamatrix/static/games_sort.js
derek-keeler: Let's open an issue for (b) and call it a day on this one.

derek-keeler: Opened #176

PR #177, opened 2026-06-26
Add select-all toggles and invert platform filter to "Include"

eniklas:
## Summary

- Adds a **tri-state master checkbox** in the Users and platform group legends (checked = all on, empty = all off, dash = some on), so a whole group can be selected/cleared in one click — on both the games filter bar and the preferences page.
- **Inverts the platform control** to **"Include platforms"** with all boxes **checked by default**, which reads more naturally than the old negative "Exclude platforms".

## Implementation notes

- New `static/toggle_all.js`: generic vanilla-JS IIFE (matching `games_sort.js` conventions, delegated listener that survives HTMX swaps). A bulk toggle dispatches a single bubbling `change` so HTMX refetches the table just once.
- The Include→exclude inversion happens **only at the request boundary** (`games/web.py` `parse_options` and `preferences.py` `save_preferences`). The stored `exclude_platforms` pref key, `ComparisonQuery`, and the comparison service are unchanged — **no data migration**, existing saved preferences keep working.
- Default and modern template variants are kept byte-identical.
- Reworked the platform tests in `tests/test_routes.py` for the inverted semantics (all-checked → exclude nothing; none-checked → exclude all; partial → exclude the rest).

## Testing

- `just check` (black, flake8, mypy, pytest): **162 passed**.

### DIFF
--- a/src/gamatrix/games/web.py
+++ b/src/gamatrix/games/web.py
@@ -14,6 +14,7 @@ from typing import Literal
 
 from fastapi import Request
 
+from gamatrix.constants import PLATFORMS
 from gamatrix.games.preferences import merge_preferences
 from gamatrix.games.service import (
     ComparisonDataset,
@@ -73,11 +74,15 @@ def parse_options(
             return qp[name] in ("true", "on", "1")
         return False if form_submitted else default
 
-    def multi(name: str, default: list[str]) -> list[str]:
-        values = qp.getlist(name)
-        if values or form_submitted:
-            return values
-        return default
+    # Platforms are shown as an inclusive list ("Include platforms") with every
+    # box checked by default; translate the included set the form returns into
+    # the excluded set the comparison service expects. On a bare load, treat
+    # everything except the saved excludes as included.
+    if form_submitted:
+        included = set(qp.getlist("include"))
+    else:
+        included = set(PLATFORMS) - set(prefs["exclude_platforms"])
+    exclude_platforms = [p for p in PLATFORMS if p not in included]
 
     # User selection: checked `user` boxes win. On a bare load fall back to the
     # saved preference, expanding the "all" sentinel to every known user.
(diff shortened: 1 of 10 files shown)


derek-keeler (commented):
I like the improvement in readability (my head doesn't do well with negatives) and bulk edit is a nice touch. I've not played with it yet, but it would make the user go through less clicks to center in on a user or two.


ChatGPT found a potential issue for you to consider... (Sorry no inline comments, not at my desk)...


Bulk toggle likely sends two HTMX requests, with a race risk.

The master checkbox’s native change event will bubble first, then toggle_all.js dispatches another bubbling change from the fieldset after changing child boxes. That can produce one stale request and one correct request; whichever finishes last may win. The PR comment explicitly relies on “one” change dispatch, but the code does not stop the original event. 

Fix proposal: call event.stopPropagation() before mutating boxes, or use a custom event for HTMX.

eniklas:
Good catch — that's a real bug, thanks (and thanks ChatGPT). Confirmed the mechanism:

HTMX's `change` listener lives on the `#filters` form, which is an **ancestor** of the master checkbox. The master's native `change` bubbles up and hits the form **before** my `document`-level listener runs — so HTMX fired request #1 with the child boxes *not yet* flipped (stale state), and then my handler flipped them and dispatched a second `change` → request #2. Race exactly as described.

A bubble-phase `stopPropagation()` wouldn't fix it (the form has already handled the event by the time my `document` listener runs). So in d8d15ff I moved the delegated listener to the **capture phase** (`addEventListener(..., true)`) so it runs *before* the form's listener, and call `event.stopPropagation()` on the master's native event. Now only the single fresh `change` dispatched after the boxes are flipped reaches HTMX — one request, correct state.

Member-checkbox changes are untouched (no `stopPropagation`), so individual toggles still go straight through to HTMX as before.

derek-keeler (approved):
Looks good! Confirmed the capture-phase listener plus event.stopPropagation() correctly prevents the master checkbox’s stale native change from reaching the HTMX form listener, while the bubbling change from the fieldset should still trigger the single fresh request.

Verification:
- node --check src/gamatrix/static/toggle_all.js passed.
- uv run pytest tests/test_routes.py -q --no-cov passed: 14 passed.

PR #179, opened 2026-06-26
Remove plain-text dialog in passkey revoke flow

derek-keeler:
## What

Fix for #163 - removes the plain-text dialog for entering your password when deleting a passkey.

## How

Replaces the plain-text dialog popup that requests your current password prior to allowing you to delete your passkey.

- Remove plain-text popup requesting users password from passkey revokation flow
  - Replace the use of window.prompt in revokation flow
  - Co-authored by codex (GPT 5.5)

- Update to handle multiple templates
  - Give the authenticated template env a fallback search path to the shared
    top-level templates dir (mode dir still wins for overrides), and convert
    default/passkeys.html.jinja to include the shared list partial like modern.
  - Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

- Remove orphaned DELETE /passkeys/{credential_id} route
  - `DELETE /auth/passkeys/{credential_id}` endpoint (and its
    DeletePasskeyRequest model) was only ever called by the window.prompt
    handler in passkeys.js
  - Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

- Fix up tests
  - the wrong/right-password deletion assertions are already covered by
      test_inline_passkey_deletion_uses_password_field_and_refreshes_list, so
      trim them from the listing test (renamed test_listing_hides_key_material)
  - repoint the two explicit-404 tests at GET /passkeys/{id}/delete
  - Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

## Checklist

- [ ] PR checks pass

### DIFF
--- a/src/gamatrix/auth/routes.py
+++ b/src/gamatrix/auth/routes.py
@@ -34,10 +34,6 @@ class CeremonyVerificationRequest(BaseModel):
     credential: dict
 
 
-class DeletePasskeyRequest(BaseModel):
-    password: str
-
-
 class CreateTokenRequest(BaseModel):
     password: str
     name: str
@@ -245,31 +241,6 @@ def passkey_authentication_verify(
         value=service.create_session_token(user["email"]), **_cookie_kwargs()
     )
     return response
-
-
-@router.delete("/passkeys/{credential_id}")
-def delete_passkey(
-    credential_id: str,
-    body: DeletePasskeyRequest,
-    user: dict = Depends(current_user_api),
-    repo: Repository = Depends(get_repo),
-):
-    if not service.authenticate(repo, user["email"], body.password):
-        raise HTTPException(
-            status_code=status.HTTP_403_FORBIDDEN, detail="Wrong password."
-        )
-    user_handle = user.get("webauthn_user_id")
-    if not user_handle:
-        raise HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
-            detail="No passkeys are registered for this account.",
-        )
-    if not repo.delete_passkey(credential_id, user_handle):
-        raise HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
-            detail="Passkey not found for this account.",
-        )
-    return {"deleted": True}
 
 
 def _token_setup_snippet(token: str) -> str:
--- a/src/gamatrix/auth/routes.py
+++ b/src/gamatrix/auth/routes.py
@@ -19,7 +19,11 @@ from gamatrix.auth.service import COOKIE_NAME
 from gamatrix.config import get_settings
 from gamatrix.constants import API_TOKEN_NAME_MAX_LENGTH
 from gamatrix.storage.dynamo import Repository
-from gamatrix.templating import authenticated_template, templates
+from gamatrix.templating import (
+    authenticated_fragment,
+    authenticated_template,
+    templates,
+)
 
 router = APIRouter(prefix="/auth", tags=["auth"])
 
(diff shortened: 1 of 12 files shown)


src/gamatrix/static/style.css
```diff
@@ -140,6 +140,10 @@ table.results td.unowned { background-color: var(--unowned); }
   display: flex; gap: 1rem; align-items: center; margin: 0.75rem 0;
 }
 .passkey-list li .muted, .token-list li .muted { flex: 1; }
+.passkey-delete-confirm { align-items: flex-start !important; flex-direction: column; }
```
eniklas: **These rules land in a stylesheet the passkeys page never loads, so the new delete form is unstyled in both modes.**

`/auth/passkeys` is rendered via `authenticated_template` → the mode env, so it extends `default/base.html.jinja` / `modern/base.html.jinja`, which link `/static/templates/default/style.css` and `/static/templates/modern/style.css` respectively. Only the public/stable base (login, etc.) links the top-level `/static/style.css` edited here. The per-mode stylesheets are hand-maintained (no generator), so the `.passkey-delete-confirm` rules above are dead for the management page — the inline form (column layout, password input border, etc.) will render unstyled in the actual default/modern UI.

Add the `.passkey-delete-confirm` rules to `src/gamatrix/static/templates/default/style.css` and `src/gamatrix/static/templates/modern/style.css` (next to the existing `.passkey-list` rules).

src/gamatrix/auth/routes.py
```diff
@@ -125,6 +137,61 @@ def list_passkeys(
     )
 
 
+@router.get("/passkeys/manage/list", response_class=HTMLResponse)
+def manage_passkeys_list(
+    request: Request,
+    user: dict = Depends(current_user_api),
+    repo: Repository = Depends(get_repo),
+):
+    return templates.TemplateResponse(
```
eniklas: **Consistency:** these new HTMX fragment routes (`manage_passkeys_list`, `confirm_delete_passkey`, `delete_passkey_inline`) render via the stable top-level `templates` env, while the page itself uses `authenticated_template` (the mode-aware env). The `templating.py` change in this PR adds the `TEMPLATES_DIR` fallback specifically so "the mode dir still wins for overrides" — but because these routes bypass the mode env, a mode that overrode `passkeys_manage_list.html.jinja` / `passkey_delete_form.html.jinja` would get its override on the initial page render but the top-level version on every HTMX refresh/confirm/error response. Today the partials only exist top-level so there's no visible divergence, but rendering these through `authenticated_fragment` (or the mode env) would keep the override contract consistent and future-proof.

eniklas (commented):
Automated review (medium effort). The flow itself looks solid — ownership is checked before the password (`_owned_passkey`), wrong passwords re-render the form without echoing the value, and the HTMX retarget refreshes the list. Two issues below: one will actually break styling in the shipped UI, the other is a consistency concern.

src/gamatrix/static/style.css
derek-keeler: `src/gamatrix/templating.py`: passkey shared partials are part of the authenticated template contract.
- src/gamatrix/static/templates/default/style.css
- src/gamatrix/static/templates/modern/style.css
- tests/test_templates.py: added coverage that authenticated stylesheets include the passkey delete-confirm styling.

src/gamatrix/auth/routes.py
derek-keeler: `src/gamatrix/auth/routes.py`: passkey HTMX fragment responses now render through authenticated_fragment.

eniklas (approved):

PR #180, opened 2026-06-27
Show confirmed single-player games as 1 player

eniklas:
## Summary

IGDB returns no max-player figure for games it confirms are single-player, so those games rendered with a **blank Players column**. This auto-populates them with **Players = 1**.

The fill happens in the comparison service (`_build_game`), gated on the game being **confirmed single-player** — IGDB lists the single-player game mode and nothing multiplayer (`multiplayer` is false). Games with an **unknown** mode set stay at `0` (blank) so we never assert a count IGDB didn't confirm.

## Why the service layer

- Covers **every already-enriched game** and the `/api/games` response, not just newly-enriched ones (the IGDB client already sets `1` for the narrow exact-`[singleplayer]` case at enrichment time; this generalizes it at read time without a re-enrich).
- Sorting by Players (`SORT_KEYS["players"]`) and the subthreshold row styling both read this value, so they stay consistent.
- Explicit `metadata_overrides` still win (applied after the base value).

## Testing

- New `test_confirmed_single_player_populates_one_player`: a single-player-mode game shows `1`; an unknown-mode game stays `0`.
- `just check` (black, flake8, mypy, pytest): **162 passed**.

### DIFF
--- a/src/gamatrix/games/service.py
+++ b/src/gamatrix/games/service.py
@@ -15,6 +15,7 @@ from typing import Literal, Protocol
 from gamatrix.config import Settings, get_settings
 from gamatrix.constants import (
     ENRICHMENT_PENDING,
+    IGDB_GAME_MODE,
     IGDB_MULTIPLAYER_GAME_MODES,
     PLATFORMS,
 )
@@ -152,7 +153,7 @@ def _build_game(rk: str, slot: dict, meta: dict, overrides: dict) -> ComparisonI
         platforms=[slot["platform"]],
         owners=sorted(slot["owners"]),
         installed=sorted(slot["installed"]),
-        max_players=meta.get("max_players", 0),
+        max_players=_display_max_players(meta),
         multiplayer=meta.get("multiplayer", False),
         rating=meta.get("rating", 0),
         rating_count=meta.get("rating_count", 0),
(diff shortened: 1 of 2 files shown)


eniklas:
A detail Claude provided in the CLI:

> One judgment call worth flagging: I treated "confirmed single-player" as the app's own classification (multiplayer == false + a single-player mode present). This means a game IGDB tags as co-op/splitscreen-without-a-count — which this app already treats as non-multiplayer — also gets 1.

I went with that since it's existing behaviour and we're a WAN-only gang, but open to discussing.

derek-keeler:
I had Codex have a look at this one (GPT-5.5) and this is what it came back with:

This should probably require the IGDB mode set to be single-player-only.

e.g. `set(game_modes) == {IGDB_GAME_MODE["singleplayer"]}`

As written, mixed modes such as `[singleplayer, coop]` or `[singleplayer, splitscreen]` can be displayed as 1 player because our stored `multiplayer` flag does not treat coop/splitscreen modes as multiplayer unless max-player data exists.

...but I am not 100% sure this is accurate? The fact that splitscreen is _technically_ multiplayer, it doesn't really add to our current use case...

eniklas:
> I had Codex have a look at this one (GPT-5.5) and this is what it came back with:
> 
> This should probably require the IGDB mode set to be single-player-only.
> 
> e.g. `set(game_modes) == {IGDB_GAME_MODE["singleplayer"]}`
> 
> As written, mixed modes such as `[singleplayer, coop]` or `[singleplayer, splitscreen]` can be displayed as 1 player because our stored `multiplayer` flag does not treat coop/splitscreen modes as multiplayer unless max-player data exists.
> 
> ...but I am not 100% sure this is accurate? The fact that splitscreen is _technically_ multiplayer, it doesn't really add to our current use case...

Yeah, see my previous comment about that. That's a judgment call Claude made that I agree with. It's technically not accurate but it's better for our WAN-only use case.

IGDB has boolean fields for `Single player`, `Multiplayer`, `Co-operative`, and `Split screen`. We only consider it single player if `Single player` is set and `Multiplayer` is not. It's ambiguous what it means to have co-op/split screen game set and Multiplayer not set, but I would guess it generally means local play only. I think it might be nice to make that distinction as part of https://github.com/eniklas/gamatrix/issues/113 though; I'll update that issue to consider this.

derek-keeler (approved):
I can dig it. 🚢 It.

PR #184, opened 2026-07-07
Chunk enrichment jobs so full refreshes don't time out

eniklas:
## Problem

Investigating why a full IGDB refresh showed **1435** in the progress bar while the library listed **2562** games, I traced the live job through CloudWatch. The enricher enqueued **every release key as a single SQS message**, so one Lambda invocation had to enrich the whole library:

- The observed job ran **900,000 ms → `Status: timeout`** (the 15-min Lambda wall) after ~466 games, and only finished because SQS **redelivered** it for a second ~8-min run.
- `completed_count` was stuck at **466/1435** because absolute progress didn't survive the timeout + redelivery.
- A genuine 2954-game **"Refresh all IGDB"** would need ~6 fifteen-minute slices but `maxReceiveCount` is 3 — so it would exhaust retries and land in the **DLQ with games left unenriched**.

(The 2562 vs 1435 gap itself was expected — 2562 is distinct titles; the job was the pending/stale subset. The real bug is that even that job couldn't complete cleanly.)

## Fix

**Chunk each job into fixed-size batches** (`ENRICHMENT_CHUNK_SIZE = 200`), one SQS message per chunk, so every invocation processes only its slice and finishes well inside the timeout; SQS fans chunks out across invocations.

- Progress is tracked per chunk in a `chunk_progress` map — each chunk writes its **own key**, so parallel chunks don't clobber each other, and per-chunk writes stay **absolute** (preserves the #131 idempotency guarantee across redelivery). `completed_count` is derived from the sum.
- The job is marked complete when the chunk that accounts for the final keys observes the sum reaching `total`. Keys needing no work (missing/already-enriched) still count, so the bar reaches 100%.
- **Cap the SQS event source at `max_concurrency=2`** so the fan-out doesn't multiply per-invocation IGDB rate limiters into a global throttle (which would surface as spurious `not_found`s).

**Also fixed `Refresh all IGDB` itself:** it re-fetched each game with a per-row `get_game` before re-stamping (≈2× the round-trips) in a serial loop inside the web request. It now re-stamps directly from the scan results via a batched `mark_games_pending`.

The **local worker** still runs whole jobs in one pass (no SQS locally) and now survives a failing job instead of crashing the poll loop.

## Tests

- `tests/test_enricher_chunking.py` (new): per-chunk fan-out on creation, chunk slicing, completion only after the last chunk, redelivery not inflating progress, and the local-worker whole-job path.
- `tests/test_jobs.py`: replaced the `set_job_progress` test with `set_chunk_progress` idempotency + non-clobbering coverage.
- `tests/test_refresh_igdb.py`: added `Refresh all` marks-every-game-pending and queues-every-game coverage.
- `just check` green: black, flake8, mypy, 174 tests. CDK stack synthesis verified.

## Follow-up (not in this PR)

The automatic staleness path (`stale_release_keys`) queues stale-but-`done` games, but the enricher only re-fetches unset/pending games — so that path still can't refresh already-enriched games that have aged past `igdb_stale_days`. Worth flipping those to `pending` like the refresh buttons do, in a separate change.

### DIFF
--- a/scripts/local_worker.py
+++ b/scripts/local_worker.py
@@ -33,7 +33,14 @@ def main() -> None:
             pending = []
         for job in pending:
             log.info("Running job %s (%d games)", job["job_id"], job.get("total", 0))
-            asyncio.run(run_job(job["job_id"], repo))
+            try:
+                # No chunk_index: locally there's no SQS fan-out, so the worker
+                # runs the whole job in one pass.
+                asyncio.run(run_job(job["job_id"], repo))
+            except Exception:
+                # One bad job shouldn't kill the poll loop; the stale reaper will
+                # mark it dead so it stops blocking new enrichment.
+                log.exception("Enrichment job %s failed", job["job_id"])
         time.sleep(POLL_SECONDS)
 
 
--- a/infrastructure/cdk/gamatrix_stack.py
+++ b/infrastructure/cdk/gamatrix_stack.py
@@ -45,6 +45,11 @@ from config import DeployConfig
 PROJECT_ROOT = Path(__file__).resolve().parents[2]
 TABLE_PREFIX = "gamatrix"
 DEFAULT_EMAIL_FROM = "noreply@example.com"
+# Max enricher invocations SQS runs in parallel. The same value is handed to the
+# enricher (ENRICHER_MAX_CONCURRENCY) so it scales its IGDB call delay and the
+# workers share, rather than each consume, IGDB's 4 req/sec budget. The SQS event
+# source enforces a floor of 2, so 2 is the lowest safe value.
+ENRICHER_MAX_CONCURRENCY = 2
 
 
 def _project_version() -> str:
@@ -168,7 +173,12 @@ class GamatrixStack(Stack):
             "EnricherFn",
             "lambda-enricher",
             "handler.handler",
-            env=common_env,
+            # The enricher scales its IGDB call delay by this so parallel
+            # invocations share the rate-limit budget (see enricher.run_job).
+            env={
+                **common_env,
+                "ENRICHER_MAX_CONCURRENCY": str(ENRICHER_MAX_CONCURRENCY),
+            },
             timeout=Duration.minutes(15),
         )
         parser_fn = self._lambda(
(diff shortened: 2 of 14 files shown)


scripts/local_worker.py
```diff
@@ -33,7 +33,14 @@ def main() -> None:
...
             log.info("Running job %s (%d games)", job["job_id"], job.get("total", 0))
-            asyncio.run(run_job(job["job_id"], repo))
+            try:
+                # No chunk_index: locally there's no SQS fan-out, so the worker
+                # runs the whole job in one pass.
+                asyncio.run(run_job(job["job_id"], repo))
+            except Exception:
+                # One bad job shouldn't kill the poll loop; the stale reaper will
+                # mark it dead so it stops blocking new enrichment.
+                log.exception("Enrichment job %s failed", job["job_id"])
```
derek-keeler, line 43: Not sure what the stale reaper is? I would have thought we would mark these jobs that cause exceptions as 'failed' (or 'done', at least, with a log entry). I haven't read too deeply into the code on how failure gets cleaned up so I'll leave this as something to be super sure about (perhaps add a comment about what the 'stale reaper' is.)

derek-keeler (approved):
Looks good, and will definitely be a performance booster for those pesky IGDB updates.

I left one inline comment about 'dead jobs' that I couldn't fully understand, but the rest looks fine.

scripts/local_worker.py
eniklas: It wasn't clear to me either, Claude says:

What it does

An enrichment job record has a status and a updated_at timestamp that the enricher bumps on each chunk of progress (enricher.py:43, and the chunk-progress write at dynamo.py:340). A job is considered stale when it's still in an active state (pending/running) but hasn't made progress within JOB_TIMEOUT_MINUTES — measured from updated_at, falling back to created_at if it never made progress.

fail_stale_jobs() scans the jobs table, and for every active-but-stale job it flips the status to JOB_FAILED and stamps completed_at, returning the reaped job ids.

Why it exists

It self-heals jobs whose enricher crashed or hit the Lambda 15-minute hard timeout without ever writing a terminal status. Left alone, such a "zombie" job would:
- pin the UI progress bar forever (the frontend polls it indefinitely), and
- block new enrichment from being queued.

It's described as the application-side backstop to the SQS queue's redrive-to-DLQ policy, which only caps redeliveries and can't touch the job record itself.

How it actually runs

There's no dedicated cron/reaper loop. The reaping happens lazily, in passing, inside get_active_job() (dynamo.py:390), which is called on page loads / status polls. So every time the app looks up the current active job, it reaps any stale ones first.

Back to your file

That's why the comment in scripts/local_worker.py:41 is accurate even though the local worker itself contains no reaping logic: when a job blows up, the worker just logs and moves on, trusting that the next get_active_job() call (via the web UI) will run fail_stale_jobs() and mark the dead job failed so it stops blocking new enrichment.

One caveat worth noting: since reaping is driven by get_active_job() rather than a background timer, a stale job is only cleaned up the next time something calls that path — if nothing hits it, the job stays "active" until it does.

